PR #2, opened 2015-08-29
Loosen signature of functions in norm.jl from Float64 to Real whenever possible

andreasnoack:
Thereby also fix bug in `normlogpdf` for non `Float64` arguments causing it to call normpdf instead of normlogpdf.

I don't think we should force all inputs to end up as `Float64`s. We should only make this restriction in the inner algorithms where we have to use the precision of the floating point number.

I wanted to look into the potential cancellation error in https://github.com/JuliaStats/Distributions.jl/blob/a4568d75baa797498983ded6a1e8f0442e326b7a/src/univariate/continuous/lognormal.jl#L56 and found the bug here, but even it hadn't been there, I would just have received the `Float64` result.

### DIFF
--- a/src/distrs/norm.jl
+++ b/src/distrs/norm.jl
@@ -1,58 +1,50 @@
 # functions related to normal distribution
 
-xval(μ::Float64, σ::Float64, z::Float64) = μ + σ * z
-zval(μ::Float64, σ::Float64, x::Float64) = (x - μ) / σ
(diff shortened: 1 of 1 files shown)


andreasnoack:
I've update the pr. Instead of multiplying with `0.5` I now divide with `2`. There should be no penalty for this because LLVM optimizes it to `0.5` again and it has the advantage that we avoid promotion to `Float64` because of `0.5`.

PR #5, opened 2015-10-21
relax restriction to Real input (allows for automatic differentiation)

mlubin:
Following the discussion in https://github.com/JuliaStats/Distributions.jl/issues/423, don't be too picky about the input types to these functions, just duck type it. This allows for AD. I left the restrictions to Real on the distribution parameters (for now)
CC @scidom @jrevels

### DIFF
--- a/src/distrs/norm.jl
+++ b/src/distrs/norm.jl
@@ -1,35 +1,35 @@
 # functions related to normal distribution
 
-xval(μ::Real, σ::Real, z::Real) = μ + σ * z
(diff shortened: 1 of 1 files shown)


scidom:
Thanks @mlubin, this one looks good to me. In principle, some of these StatsFun functions would allow us to run the example(s) we have in mind without having to necessarily use Distributions.

mlubin:
Thanks!

mlubin:
This can basically be reverted now that ForwardDiff numbers are subtypes of Real.

PR #6, opened 2016-03-29
generalize logmvgamma to handle real a

jmxpearson:
- needed to work with automatic differentiation of wishart

### DIFF
--- a/src/misc.jl
+++ b/src/misc.jl
@@ -5,7 +5,7 @@
 # See: https://en.wikipedia.org/wiki/Multivariate_gamma_function
 #
 function logmvgamma(p::Int, a::Real)
-    # NOTE: one(a) factors are here for type stability for, e.g., Float32
+    # NOTE: one(a) factors are here to prevent unnecessary promotion of Float32
     res = p * (p - 1) * log(pi * one(a)) / 4
     for ii in 1:p
         res += lgamma(a + (1 - ii) * one(a)/ 2)
--- a/src/misc.jl
+++ b/src/misc.jl
@@ -5,9 +5,10 @@
 # See: https://en.wikipedia.org/wiki/Multivariate_gamma_function
 #
 function logmvgamma(p::Int, a::Real)
-    res = p * (p - 1) / 4 * log(pi * one(a))
+    # NOTE: one(a) factors are here for type stability for, e.g., Float32
+    res = p * (p - 1) * log(pi * one(a)) / 4
     for ii in 1:p
-        res += lgamma(a + (1 - ii) / 2)
+        res += lgamma(a + (1 - ii) * one(a)/ 2)
     end
     return res
 end
--- a/src/misc.jl
+++ b/src/misc.jl
@@ -5,9 +5,9 @@
 # See: https://en.wikipedia.org/wiki/Multivariate_gamma_function
 #
 function logmvgamma(p::Int, a::Real)
-    res = p * (p - 1.0) / 4.0 * log(pi)
+    res = p * (p - 1) / 4 * log(pi * one(a))
     for ii in 1:p
-        res += lgamma(a + (1.0 - ii) / 2.0)
+        res += lgamma(a + (1 - ii) / 2)
     end
     return res
 end
--- a/src/misc.jl
+++ b/src/misc.jl
@@ -4,8 +4,8 @@
 #
 # See: https://en.wikipedia.org/wiki/Multivariate_gamma_function
 #
-function logmvgamma(p::Int, a::Float64)
-    res::Float64 = p * (p - 1.0) / 4.0 * log(pi)
+function logmvgamma(p::Int, a::Real)
+    res = p * (p - 1.0) / 4.0 * log(pi)
     for ii in 1:p
         res += lgamma(a + (1.0 - ii) / 2.0)
     end


src/misc.jl
```diff
@@ -4,8 +4,8 @@
 #
 # See: https://en.wikipedia.org/wiki/Multivariate_gamma_function
 #
-function logmvgamma(p::Int, a::Float64)
-    res::Float64 = p * (p - 1.0) / 4.0 * log(pi)
+function logmvgamma(p::Int, a::Real)
+    res = p * (p - 1.0) / 4.0 * log(pi)
```
andreasnoack: This is actually a case where the body exploits `Float64` by using `Float64` literals. However, I think that it's fine to use integers and `div` instead of `/`. That would make the function more generic. Maybe move `p=1` out of the loop to make sure that the type of `res` is stable over the loop.

jmxpearson: I think I'm missing something.

1. Doesn't `log(pi)` on line 8 mean that res is always *at least* as big as `Float64`?
1. Where is the type instability again? If there's a cast, shouldn't it be
```julia
res::typeof(a) = p * (p - 1.0) / 4.0 * log(pi)
```

andreasnoack: I was thinking of `Float32`. It seems wrong to promote to `Float64` if you allow `Real` input. You are right that `log(pi)` is `Float64` so I'm not completely sure how this should be handled. Maybe `log(pi*one(a))`?

jmxpearson: I think that does it. See changes.

src/misc.jl
```diff
@@ -4,10 +4,10 @@
 #
 # See: https://en.wikipedia.org/wiki/Multivariate_gamma_function
 #
-function logmvgamma(p::Int, a::Float64)
-    res::Float64 = p * (p - 1.0) / 4.0 * log(pi)
+function logmvgamma(p::Int, a::Real)
+    res = p * (p - 1) / 4 * log(pi * one(a))
```
andreasnoack: Since `/` is floating point division I think it is necessary to move the division to after the `log`.

src/misc.jl
```diff
@@ -4,10 +4,10 @@
...
     for ii in 1:p
-        res += lgamma(a + (1.0 - ii) / 2.0)
+        res += lgamma(a + (1 - ii) / 2)
```
andreasnoack: ...but this one is tricky. `(1 - ii)` would have to be converted to `typeof(a)` before the division.

jmxpearson: Like `(1 - ii) * one(a)/2`?

andreasnoack: That would probably work. It would be great if you could add a small comment explaining why these weird `one`s are around.

jmxpearson:
Done. Take a look at this, and if it seems good, I'll squash.

src/misc.jl
```diff
@@ -4,10 +4,11 @@
 #
 # See: https://en.wikipedia.org/wiki/Multivariate_gamma_function
 #
-function logmvgamma(p::Int, a::Float64)
-    res::Float64 = p * (p - 1.0) / 4.0 * log(pi)
+function logmvgamma(p::Int, a::Real)
+    # NOTE: one(a) factors are here for type stability for, e.g., Float32
```
andreasnoack: Instead of "type stability", I'd say something like "avoid unnecessary promotion of e.g.  `Float32`"

src/misc.jl
```diff
@@ -4,10 +4,11 @@
...
+    res = p * (p - 1) * log(pi * one(a)) / 4
```
andreasnoack:
```jl
julia> oftype(Int, pi)
WARNING: oftype{T}(::Type{T},c) is deprecated, use convert(T,c) instead.
 [inlined code] from ./error.jl:26
 in depwarn(::ASCIIString, ::Symbol) at ./deprecated.jl:64
 in oftype(::Type{Int64}, ::Irrational{:π}) at ./deprecated.jl:50
 in eval(::Module, ::Any) at /Users/andreasnoack/julia-dev/usr/lib/julia/sys.dylib:-1
while loading no file, in expression starting on line 0
ERROR: MethodError: Cannot `convert` an object of type Irrational{:π} to an object of type Int64
This may have arisen from a call to the constructor Int64(...),
since type constructors fall back to convert methods.
Closest candidates are:
  convert(::Type{Int64}, ::Int8)
  convert(::Type{Int64}, ::Int16)
  convert(::Type{Int64}, ::Int32)
  ...
 in oftype(::Type{Int64}, ::Irrational{:π}) at ./deprecated.jl:52
 in eval(::Module, ::Any) at /Users/andreasnoack/julia-dev/usr/lib/julia/sys.dylib:-1

julia> one(Int)*pi
3.141592653589793
```

jmxpearson: Sorry, missed this last comment. 

Keep `one(a)` or not?

andreasnoack: Keep. We'll have to rely on the promoting rules here.

nalimilan: Ah, this is really tricky...

jmxpearson:
Bump

johnmyleswhite:
I think is good to merge. @jmxpearson: I tried to give you commit access recently, so feel free to merge on your own.

jmxpearson:
@johnmyleswhite: Hmmm. I don't see a merge button and merge->push from the command line failed. Could you double-check?

johnmyleswhite:
I retried the invite.

jmxpearson:
Ugh. Forgot to squash. Apologies.

johnmyleswhite:
It's ok. I'm sure you'll remember next time. I hope one day GitHub makes the merging tools better since the squashing should be more automated.

johnmyleswhite:
Sounds like the relevant functionality should be available soon: https://github.com/blog/2141-squash-your-commits

jmxpearson:
Just saw this. Nice.

PR #8, opened 2016-05-12
Prepare for 0.5. Use Symbol instead of symbol

andreasnoack:


tkelman:
Add a minimum Compat version requirement when you do these.

nalimilan:
> Add a minimum Compat version requirement when you do these.

I just pushed a commit to fix this.

PR #21, opened 2016-11-07
Adding pure Julia pdf and logpdf as callbacks.

xukai92:
This PR is aimed to resolve the incompatibility between `Distributions.jl` package and `ForwardDiff.jl` package. 

As `Distributions.jl` is dependent on `StatsFuns.jl` and `StatsFuns.jl` imports `Rmath.jl`, we cannot pass `Dual` type numbers to distributions in the use of the automatic differentiation technique. This PR implements the pdf and logpdf of Beta and Gamma to solve this issues about them.


src/distrs/gamma.jl
```diff
@@ -11,3 +9,10 @@ import .RFunctions:
     gammainvccdf,
     gammainvlogcdf,
     gammainvlogccdf
+
+
+# pdf
+gammapdf(k::Real, θ::Real, x::Number) = 1 / (gamma(k) * θ^k) * x^(k - 1) * exp(-x / θ)
+
+# logpdf
+gammalogpdf(k::Real, θ::Real, x::Number) = -log(gamma(k)) - k * log(θ) + (k - 1) * log(x) - x / θ
```
ararslan: `lgamma(k)` would probably be better than `log(gamma(k))` here

src/distrs/beta.jl
```diff
@@ -12,5 +11,7 @@ import .RFunctions:
     betainvlogccdf
 
 # pdf
-betapdf(α::Float64, β::Float64, x::Float64) = exp(betalogpdf(α, β, x))
-betapdf(α::Real, β::Real, x::Real) = betapdf(Float64(α), Float64(β), Float64(x))
+betapdf(α::Real, β::Real, x::Number) = x^(α - 1) * (1 - x)^(β - 1) / beta(α, β)
+
+# logpdf
+betalogpdf(α::Real, β::Real, x::Number) = (α - 1) * log(x) + (β - 1) * log(1 - x) - log(beta(α, β))
```
ararslan: I think we have an `lbeta` function in Base that should be better than `log(beta())`

xukai92: Thanks! I actually didn't know there are `lgamma` and `lbeta`. Just changed them.

src/distrs/beta.jl
```diff
@@ -12,5 +11,7 @@ import .RFunctions:
     betainvlogccdf
 
 # pdf
-betapdf(α::Float64, β::Float64, x::Float64) = exp(betalogpdf(α, β, x))
-betapdf(α::Real, β::Real, x::Real) = betapdf(Float64(α), Float64(β), Float64(x))
+betapdf(α::Real, β::Real, x::Number) = x^(α - 1) * (1 - x)^(β - 1) / beta(α, β)
+
+# logpdf
+betalogpdf(α::Real, β::Real, x::Number) = (α - 1) * log(x) + (β - 1) * log(1 - x) - lbeta(α, β)
```
ararslan: Any idea whether `log1p(-x)` would be more efficient than `log(1 - x)`? cc @simonbyrne, who knows many things

xukai92: Thank you for cc Simon.

BTW I did a stupid/quick check by `@elapsed [log1p(-x) for _ in 1:200000000]` versus `@elapsed [log(1-x) for _ in 1:200000000]` with `x = -1.2` set, which gives me `17.890346949`s for `log1p()` and `16.294338502`s for `log()` on my local machine. Does this way for checking make sense or I need to do check on random `x`s?

ararslan: In generally I usually prefer using the BenchmarkTools package for things like this. I did a quick benchmark and they appear to give basically identical results. It's looking like `log1p` probably isn't necessary and that what you have is perfectly fine as-is.

andreasnoack: `log1p` is for precision. The number density around zero is much higher than one and `log1p` maps a small neighborhood of zero to a small neighborhood of zero.
```
julia> log1p(eps()/2)
1.1102230246251565e-16
```
vs
```
julia> log(1 + eps()/2) # (= log(1.0))
0.0
```
Before we can consider this PR, we'd have to check the precision compared to the implementations in Rmath.

xukai92: All right. Thanks!

simonbyrne:
Thanks for having a stab at this. I actually had a shot at this a long time ago (based on some code that @andreasnoack had translated) here: https://github.com/JuliaStats/Distributions.jl/pull/138

This probably makes sense as a fallback, but you might still hit some boundary cases in general usage (e.g. the log-space function will probably see catastrophic cancellation, and the `pdf` might have intermediate underflow)

As an example, the core kernel function that we used on the aforementioned PR (based on the old NSWC fortran code) was this one:
https://github.com/JuliaStats/Distributions.jl/blob/91366c9aab6f6497b0d100eb41321eb4cb0668c5/src/specialfuns/gammabeta.jl#L2330-L2413

or you can look at Boost: https://github.com/boostorg/math/blob/242e9d669198415d165bb0213f730c00985bee9a/include/boost/math/special_functions/beta.hpp#L1394

One option is to keep the existing `Rmath` definitions for `Float64`, and use the new ones as fallbacks?

(also, on that note, why isn't the `betapdf` directly calling `Rmath.betapdf`?)

andreasnoack:
> One option is to keep the existing `Rmath` definitions for `Float64`, and use the new ones as fallbacks?

I think this a good way to get this moving. The implementation should still use `log1p` though.

src/distrs/beta.jl
```diff
@@ -1,7 +1,6 @@
 # functions related to beta distributions
 
 import .RFunctions:
-    betalogpdf,
```
andreasnoack: You don't want to remove this one and you probably want to add `betapdf`. Same with `gamma`. The new definitions should only be for non `Float64` elements.

ararslan: I'm not sure what you mean when you say it won't overload, but if `Real` works and `Number` doesn't, I'd guess that it's because `Real` is more specific than `Number` so it's not overwriting an existing method if one for `Number` exists.

xukai92: Yes you are right. I guess in order to use the original `Rmath` functions for `Float64` and my code for other generic types, I also need to change the original definition of the imports in`rmath.jl` from `Real` to `Float64`. Is my understanding right?

xukai92: I start thinking it should be recognized as a bug of `ForwardDiff.jl` because the `Dual` type in their package inherits from `Real` but actually the true relation should be `Dual` extents `Real`.

src/rmath.jl
```diff
@@ -53,10 +53,10 @@ function _import_rmath(rname::Symbol, jname::Symbol, pargs)
 
     # function implementation
     quote
-        $pdf($(pdecls...), x::Real) =
+        $pdf($(pdecls...), x::Float64) =
```
andreasnoack: I think it is the right thing to do. Conversion shouldn't happen in the function that `ccall`s. Also, I don't think `StatsFuns` should do much conversion at all so you can just go ahead and change the tests in https://github.com/JuliaStats/StatsFuns.jl/blob/master/test/rmath.jl#L123-L129 to use floats instead of integers.

andreasnoack:
Almost there. @xukai92 Would you mind to try out the Distributions` tests against this branch of `StatsFuns`. Just so we know if anything breaks.

xukai92:
@andreasnoack I tested it and it turned out that `Distributions` uses tests which pass `Int` to the function, e.g. `binomlogpdf`. I guess we also need provide the `Int` interface to use `Rmath` in our side? Because ideally we only want `Dual` to use our hand-written `pdf` and `logpdf` functions. Does it make sense?

xukai92:
@andreasnoack `Distributions`'s tests pass on my local machine now.

xukai92:
I also want to add more fallbacks functions to other distributions functions. Do you think it's better to do those in this PR or better after this PR?

xukai92:
What is the failure on test? I can't tell why/.

src/rmath.jl
```diff
@@ -56,9 +56,15 @@ function _import_rmath(rname::Symbol, jname::Symbol, pargs)
...
 
+        $pdf($(pdecls...), x::Int64) =
+            ccall(($dfun, libRmath), Float64, $dtypes, x, $(pargs...), 0)
+
         $logpdf($(pdecls...), x::Float64) =
             ccall(($dfun, libRmath), Float64, $dtypes, x, $(pargs...), 1)
 
+        $logpdf($(pdecls...), x::Int64) =
+            ccall(($dfun, libRmath), Float64, $dtypes, x, $(pargs...), 1)
+
```
ararslan: Rather than having separate signatures for `Float64` and `Int64`, you should just be able to do `x::Union{Float64,Int64}`.

xukai92: I see! Thank you!

andreasnoack:
The error is unrelated. It is probably because of the deprecation of `$` for xor in Base.

ararslan:
I restarted the failed nightly job just in case but I agree that it's almost certainly unrelated.

xukai92:
I implemented more functions in generic way

- I didn't do any noncentral one because it seems that `bessel` is not pure Julia implemented.
- I added some tests which uses `ForwardDiff.jl`'s `Dual` - is it allowed?

xukai92:
> - I added some tests which uses `ForwardDiff.jl`'s `Dual` type - is it allowed to use another package in the test file?
It seems that I can't use `ForwardDiff.jl` in the test file. Can you indicate me way to make it possible or the tests for generic callbacks somehow?

andreasnoack:
@xukai92 You can add a `REQUIRE` file to the `test` directory with test dependencies. I think it would be fine to add `ForwardDiff` as a test dependency.

xukai92:
@andreasnoack Thank you for pointing this! Now the tests are passed.

src/distrs/chisq.jl
```diff
@@ -11,3 +11,15 @@ import .RFunctions:
     chisqinvccdf,
     chisqinvlogcdf,
     chisqinvlogccdf
+
+# pdf for numbers with generic types
+function chisqpdf(k::Real, x::Number)
+  hk = 0.5 * k  # hlaf k
```
andreasnoack: Don't use `Float64` literals. Divide with `2` instead. The compiler will make it as efficient.

src/distrs/chisq.jl
```diff
@@ -11,3 +11,15 @@ import .RFunctions:
...
+  1 / (2^(hk) * gamma(hk)) * x^(hk - 1) * exp(-0.5 * x)
+end
+
+# logpdf for numbers with generic types
+function chisqlogpdf(k::Real, x::Number)
+  hk = 0.5 * k  # hlaf k
+  -hk * log(2) - lgamma(hk) + (hk - 1) * log(x) - 0.5 * x
```
andreasnoack: Use `log(oftype(hk, 2))` here. Otherwise `log(2)` will promote `Float32` to `Float64` because `log(2)` will be `Float64`.

test/callbacks.jl
```diff
@@ -0,0 +1,113 @@
+using StatsFuns
+using ForwardDiff: Dual
+
```
andreasnoack: Why do you call this file `callbacks.jl`? Isn't it mainly to test that generic types exemplified by `Dual`s work?

xukai92:
I've changed the code following your comments.

src/distrs/chisq.jl
```diff
@@ -11,3 +11,15 @@ import .RFunctions:
...
+# pdf for numbers with generic types
+function chisqpdf(k::Real, x::Number)
+  hk = k / 2  # hlaf k
+  1 / (2^(hk) * gamma(hk)) * x^(hk - 1) * exp(-0.5 * x)
+end
+
+# logpdf for numbers with generic types
+function chisqlogpdf(k::Real, x::Number)
+  hk = k / 2  # hlaf k
+  -hk * log(oftype(hk, 2)) - lgamma(hk) + (hk - 1) * log(x) - 0.5 * x
```
andreasnoack: There are several places where you use `0.5`. They should all be changed.

xukai92: All right.

andreasnoack:
@xukai92 Thanks for this PR. It's a great feature to be able to use these distributions in automatic differentiation.

xukai92:
@andreasnoack You're welcome! Very glad to see this PG got accepted.

test/rmath.jl
```diff
@@ -185,8 +185,8 @@ rmathcomp_tests("norm", [
     ((0.0, 1.0), -6.0:0.01:6.0),
     ((2.0, 1.0), -3.0:0.01:7.0),
     ((0.0, 0.5), -3.0:0.01:3.0),
-    ((0, 1), -3:3),
-    ((0, 1), -Float16(3):Float16(0.01):Float16(3))
```
tkelman: why did this remove testing for float16?

andreasnoack: As I remember it, the reason was that we decided that Rmath should only deal with `Float64` and not automatically convert inputs. One reason was the conversions should be handled explicitly such that we didn't end up promoting `Float32` to `Float64` by accident but restricting some of the Rmath signatures also made it easier to handle dispatch to generic implementations which was the main purpose of this PR.

I just checked and the `Float16` tests were not really that useful since in 0.5 multiplication with `2` converted them to `Float32`s so we don't really handle `Float16`s right now. Adding the tests back would require that we more seriously handle `Float16` inputs but that is work for another day.

tkelman: Isn't tightening a lot of the signatures potentially breaking?

ararslan: Potentially but unlikely. This was tagged in a minor rather than a patch release, right?

tkelman: patch but not merged yet https://github.com/JuliaLang/METADATA.jl/pull/7590

tkelman: integer literals on 32 bit systems would also go to a different code path here - why was this using Int64 rather than Int?

src/rmath.jl
```diff
@@ -53,10 +53,10 @@ function _import_rmath(rname::Symbol, jname::Symbol, pargs)
 
     # function implementation
     quote
-        $pdf($(pdecls...), x::Real) =
+        $pdf($(pdecls...), x::Union{Float64,Int64}) =
             ccall(($dfun, libRmath), Float64, $dtypes, x, $(pargs...), 0)
 
-        $logpdf($(pdecls...), x::Real) =
+        $logpdf($(pdecls...), x::Union{Float64,Int64}) =
```
tkelman: specifically I'm referring to these lines - Int64 doesn't seem appropriate here

test/rmath.jl
andreasnoack: I see. I've opened https://github.com/JuliaStats/StatsFuns.jl/pull/26. When that has been merged, I'll move the tag and make it a minor.

PR #23, opened 2016-12-03
Added handling of Infs and NaNs to logsumexp, with tests.

### LINKED ISSUE #22: logsumexp([-Inf,-Inf]) incorrect (gives NaN)
The correct value would be `-Inf`. The bug occurs because the maximum is `-Inf`, resulting in a `-Inf - (-Inf)==NaN`. Catching `u == -Inf` and returning `-Inf` is a possible solution.

Two-argument `logsumexp` has the same problem.

Unless the maintainers prefer to make this trivial change, I can make up a PR (with tests).

tpapp:
This version avoids the `Inf-Inf` errors which (incorrectly) resulted in
`NaN`s before, and also takes care to pass through actual `NaN`s.

New tests for both vector and bivariate forms.

Fixes #22.

### DIFF
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -129,8 +129,8 @@ logsumexp(x::Real, y::Real) = logsumexp(promote(x, y)...)
 function logsumexp{T<:Real}(x::AbstractArray{T})
     isempty(x) && return -T(Inf)
     u = maximum(x)
-    abs(u) == Inf && return any(isnan.(x)) ? T(NaN) : u
-    s = zero(T)
+    abs(u) == Inf && return any(isnan(x)) ? T(NaN) : u
+    s = zero(Base.promote_op(exp, T))
     for i = 1:length(x)
         @inbounds s += exp(x[i] - u)
     end
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -119,13 +119,18 @@ end
 
 ## logsumexp
 
-logsumexp{T<:Real}(x::T, y::T) = x > y ? x + log1p(exp(y - x)) : y + log1p(exp(x - y))
+function logsumexp{T<:Real}(x::T, y::T)
+    x == y && abs(x) == Inf && return x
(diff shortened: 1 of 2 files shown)


src/basicfuns.jl
```diff
@@ -119,13 +119,18 @@ end
...
+end
+
 logsumexp(x::Real, y::Real) = logsumexp(promote(x, y)...)
 
 function logsumexp{T<:Real}(x::AbstractArray{T})
-    isempty(x) && return -Inf
+    isempty(x) && return -T(Inf)
     u = maximum(x)
-    s = 0.
+    abs(u) == Inf && return any(isnan.(x)) ? T(NaN) : u
```
andreasnoack: Use `any(isnan, x)` instead. It avoids a temporary and will also work on 0.4.

src/basicfuns.jl
```diff
@@ -119,13 +119,18 @@ end
...
+    s = zero(T)
```
andreasnoack: This is not a type stable accumulator if e.g. `T=Int`. You could do `zero(promote_op(exp,T))`.

tpapp:
@andreasnoack : Thanks for the help, made the changes you requested.

src/basicfuns.jl
```diff
@@ -119,13 +119,18 @@ end
...
+end
+
 logsumexp(x::Real, y::Real) = logsumexp(promote(x, y)...)
 
 function logsumexp{T<:Real}(x::AbstractArray{T})
-    isempty(x) && return -Inf
+    isempty(x) && return -T(Inf)
     u = maximum(x)
-    s = 0.
+    abs(u) == Inf && return any(isnan(x)) ? T(NaN) : u
```
andreasnoack: This should be `any(isnan, x)` not `any(isnan(x))`. The first version is `mapreduce`, i.e. applies `isnan` to each element before the `any` reduction`.

andreasnoack:
Thanks for the PR.

PR #24, opened 2017-01-17
Fix deprecation warnings for Julia 0.6

andreasnoack:


src/tvpack.jl
```diff
@@ -124,10 +124,10 @@ end
...
 
 function adonet(f::Function, a::Float64, b::Float64, tol::Float64)
-	ai = Array(Float64, 100)
-	bi = Array(Float64, 100)
-	ei = Array(Float64, 100)
-	fi = Array(Float64, 100)
+	ai = Array{Float64}(100)
+	bi = Array{Float64}(100)
+	ei = Array{Float64}(100)
+	fi = Array{Float64}(100)
```
ararslan: Might make it easier on the compiler to use `Vector` here so it doesn't have to determine the number of dimensions itself.

src/basicfuns.jl
```diff
@@ -148,4 +148,4 @@ function softmax!{R<:AbstractFloat,T<:Real}(r::AbstractArray{R}, x::AbstractArra
 end
 
 softmax!{T<:AbstractFloat}(x::AbstractArray{T}) = softmax!(x, x)
-softmax{T<:Real}(x::AbstractArray{T}) = softmax!(Array(Float64, size(x)), x)
+softmax{T<:Real}(x::AbstractArray{T}) = softmax!(Array{Float64}(size(x)), x)
```
ararslan: Could do `Array{Float64,ndims(x)}(size(x))`

test/basicfuns.jl
```diff
@@ -7,82 +7,82 @@ using Compat
...
-@test_approx_eq logsumexp([1.0, 2.0, 3.0] .+ 1000.) 1003.40760596444438
+@test logsumexp([1.0, 2.0, 3.0])          ≈ 3.40760596444438
+@test logsumexp([1.0, 2.0, 3.0] .+ 1000.) ≈ 1003.40760596444438
 
 # softmax
 
 println("\ttesting softmax ...")
 
 x = [1.0, 2.0, 3.0]
 r = @compat exp.(x) ./ sum(exp.(x))
```
ararslan: For whatever reason, this is inferred as `Array{Any,1}` on 0.4, which is causing test failures. It could be written as
```julia
r = map(exp, x) ./ sum(exp, x)
```
which should work across versions (and should render `@compat` unnecessary).

src/basicfuns.jl
andreasnoack: I don't think the tiny compilation saving here is worth it added verbosity.

test/basicfuns.jl
andreasnoack: With compat these translates for `broadcast` calls and on 0.4 they cannot be inferred. I've added explicit conversion temporarily. It can be removed when we drop 0.4 support.

PR #27, opened 2017-04-09
Fix test error on 0.6

yuyichao:


src/StatsFuns.jl
```diff
@@ -228,6 +228,9 @@ include("misc.jl")
 include("rmath.jl")
 
 using .RFunctions
+if VERSION >= v"0.6.0-dev.2767"
```
ararslan: The version check isn't necessary since SpecialFunctions supports 0.5.

ararslan:
Thanks as always, Yichao! While you're in here, if you don't mind, could you remove the 0.4 badge and add `[StatsFuns](http://pkg.julialang.org/?pkg=StatsFuns)` to the README after the 0.5 badge?

src/StatsFuns.jl
yuyichao: it's needed or there was a test error.

ararslan: What was the test error?

yuyichao: Conflict import. You can check CI log. Basically SpecialFunctions isn't precise about importing/extending functions.

PR #30, opened 2017-07-12
Binomial generic support

xukai92:
Support

- `binompdf(n::Real, p::Real, k::Number)`, and
- `binomlogpdf(n::Real, p::Real, k::Number)`

as call-backs functions when `k` is neither `Float64` nor `Int`


src/distrs/binom.jl
```diff
@@ -11,3 +11,21 @@ import .RFunctions:
     binominvccdf,
     binominvlogcdf,
     binominvlogccdf
+
+# pdf for numbers with generic types
+binompdf(n::Real, p::Real, k::Number) = begin
+    # NOTE: fail to do below
+    #       prod(map(i -> (n + 1 - i) / i, 1:Int(k.))) * p^k * (1 - p)^(n - k)
+    n_choose_k = 1; i = 1;
+    while i <= k n_choose_k *= (n + 1 - i) / i; i += 1 end
```
andreasnoack: This can overflow for large `n` and `k`. Probably better to compute this as `exp(binomlogpdf)`

src/distrs/binom.jl
```diff
@@ -11,3 +11,21 @@ import .RFunctions:
...
+    n_choose_k * p^k * (1 - p)^(n - k)
+end
+
+# logpdf for numbers with generic types
+binomlogpdf(n::Real, p::Real, k::Number) = begin
```
andreasnoack: Why is `k` only restricted to `Number` here?

src/distrs/binom.jl
```diff
@@ -11,3 +11,21 @@ import .RFunctions:
...
+    # NOTE: fail to do below
+    #       sum(map(i -> log((n + 1 - i) / i), 1:k)) + k * log(p) + (n - k) * log(1 - p)
+    log_n_choose_k = 0; i = 1;
+    while i <= k log_n_choose_k += log((n + 1 - i) / i); i += 1 end
+    log_n_choose_k + k * log(p) + (n - k) * log(1 - p)
```
andreasnoack: `log(1-p)` -> `log1p(p)` but the best solution would probably be to make `beta` work for `Dual` and [here](https://github.com/JuliaDiff/DiffBase.jl/blob/master/src/rules.jl) with the derivative rules https://www.wolframalpha.com/input/?i=d%2Fda+beta(a,b)

src/distrs/binom.jl
xukai92: I see!

src/distrs/binom.jl
xukai92: I just followed the convention in https://github.com/JuliaStats/StatsFuns.jl/blob/master/src/distrs/norm.jl.

src/distrs/binom.jl
xukai92: I see. For the 2nd point, do you mean it's better to implement d/da beta and d/db beta s.t. `Dual` can use them directly instead of pass things down?

andreasnoack: If beta works with dual then you can use the beta function here in the binomial pdf

andreasnoack: I decided to add rules for `beta` and `lbeta` to DiffBase to figure out how it works so you can use `lbeta` here instead of rolling your own binomial coefficient.

src/distrs/binom.jl
andreasnoack: I see. I guess it is left over from the time when `Dual`s weren't `Real`. Should probably be `Real` here.

src/distrs/binom.jl
xukai92: OK cool. Then do I need to wait for the PR to be accepted?

src/distrs/binom.jl
xukai92: Do you want me to change all of them (there are other in those call-backs I implemented before)?

src/distrs/binom.jl
andreasnoack: You can just start adjusting the PR. I'm confident it will get merged very soon.

src/distrs/binom.jl
andreasnoack: It would be great if you could do that as part of a separate PR.

xukai92:
I've made those changes required.

src/distrs/binom.jl
```diff
@@ -13,19 +13,9 @@ import .RFunctions:
...
 
 # logpdf for numbers with generic types
-binomlogpdf(n::Real, p::Real, k::Number) = begin
-    # NOTE: fail to do below
-    #       sum(map(i -> log((n + 1 - i) / i), 1:k)) + k * log(p) + (n - k) * log(1 - p)
-    log_n_choose_k = 0; i = 1;
-    while i <= k log_n_choose_k += log((n + 1 - i) / i); i += 1 end
-    log_n_choose_k + k * log(p) + (n - k) * log(1 - p)
+binomlogpdf(n::Real, p::Real, k::Real) = begin
+    -log(n + 1) - lbeta(n - k + 1, k + 1) + k * log(p) + (n - k) * log(1 - p)
```
andreasnoack: `log(n+1)` -> `log1p(n)` (less important) and `log(1-p)` -> `log1p(-p)` (quite important)

src/distrs/binom.jl
```diff
@@ -13,19 +13,9 @@ import .RFunctions:
...
+binompdf(n::Real, p::Real, k::Real) = exp(binomlogpdf(n, p, k))
 
 # logpdf for numbers with generic types
-binomlogpdf(n::Real, p::Real, k::Number) = begin
-    # NOTE: fail to do below
-    #       sum(map(i -> log((n + 1 - i) / i), 1:k)) + k * log(p) + (n - k) * log(1 - p)
-    log_n_choose_k = 0; i = 1;
-    while i <= k log_n_choose_k += log((n + 1 - i) / i); i += 1 end
-    log_n_choose_k + k * log(p) + (n - k) * log(1 - p)
+binomlogpdf(n::Real, p::Real, k::Real) = begin
```
andreasnoack: maybe just remove the `begin` and `end`

src/distrs/binom.jl
xukai92: Yes!

xukai92:
Thx for helping through :smile:

andreasnoack:
No problem. Thanks for improving the package.

andreasnoack:
Seems like we need a release of ForwardDiff

andreasnoack:
Sorry for the delay here. It seems that we also need a new version of Compat before StatsFuns work on master. https://github.com/JuliaLang/Compat.jl/pull/394

nalimilan:
A new Compat version has been merged now, should be OK to rebase and merge?

PR #31, opened 2017-12-15
Fix deprecations on 0.7

ararslan:
Also bump the minimum Julia version to 0.6


ararslan:
This doesn't actual fix 0.7 but it fixes a few deprecations so #yolo

PR #32, opened 2018-03-01
Native binompdf and binomlogpdf

Ellipse0934:
Added native binompdf and binomlogpdf functions. Small performance improvement. Tests needed.

### DIFF
--- a/src/StatsFuns.jl
+++ b/src/StatsFuns.jl
@@ -226,6 +226,7 @@ include("constants.jl")
 include("basicfuns.jl")
 include("misc.jl")
 include("rmath.jl")
+include("binomial.jl")
 
 using .RFunctions
 using SpecialFunctions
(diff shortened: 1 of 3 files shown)


Ellipse0934:
Will fix failed tests and reopen.

dmbates:
How did you formulate the algorithm?  It looks similar to the approach in the Rmath library and, technically, you can't read the Rmath code then write a version covered by a less restrictive license in another language.

Many of us would like to have native Julia versions of these functions but I think the way it will need to be done, silly as this sounds, is for one person or group to read the Rmath code and write a narrative description or some type of pseudo code and another group write the Julia version without ever looking at the Rmath code itself.

Ellipse0934:
I agree with what you have said. At octave people who would write functions were forbidden from viewing matlab source because of potential licensing issues. However if you have a look at this paper -> http://octave.1599824.n4.nabble.com/attachment/3829107/0/loader2000Fast.pdf . The crux of this algorithm is computation of the deviance and stirling error functions. There is almost no flexibility in writing these functions and the rest is straight forward. I was unsuccessful in asking the author for permission (couldn't find her contact). It's upto the reviewers.

PR #33, opened 2018-03-03
Native binompdf and binomlogpdf

Ellipse0934:
Includes tests.


src/binomial.jl
```diff
@@ -0,0 +1,93 @@
...
+```
+# Arguments
+- `n::Integer`: total number of trials.
+- `p::Float`: probability of success of each trial.
+- `x::Integer`: number of successful trials.
+...
+"""
+function binomlogpdf(n::Integer, p::AbstractFloat, x::Integer)
+    return log(binompdf(n, p, x))
+end
```
jwmerrill: It might be better to compute this directly rather than taking the log of binompdf. The implementation of binompdf already effectively computes the logarithmic form and then exponentiates it, so binomlogpdf could be implemented with a similar algorithm that just doesn't exponentiate at the end.

One reason to prefer a direct implementation is that binompdf can underflow to 0, and then taking its log will give -Inf, even though binomlogpdf could have been represented accurately.

For example, using the existing implementation:
```
julia> logpdf(Binomial(1200, 0.5), 1)
-824.6865398361583
```
but with this new implementation,
```
julia> binomlogpdf(1200, 0.5, 1)
-Inf
```

Ellipse0934:
Sorry for littering with so many fixes. I was experimenting with ways to unit test.

Ellipse0934:
In PR#35 dmbates added generic poisson pdf and logpdf functions. However, I have removed them in my PR as I didn't see a a point for their existence. Does it need to be added back ?

src/binomial.jl
```diff
@@ -0,0 +1,103 @@
+function stirlerr(n::Int64)
```
andreasnoack: This function is written for `Float64` which should probably be indicated somehow such that it is not suddenly called as part of a `BigFloat` computation. Maybe a way to do that would be to use `n::Float64` instead of `n::Int64`.

src/binomial.jl
```diff
@@ -0,0 +1,103 @@
...
+julia> binompdf(13, 0.58, 7)
+0.20797396939077062
+```
+# Arguments
+- `n::Integer`: total number of trials.
+- `p::Float`: probability of success of each trial.
+- `x::Integer`: number of successful trials.
+...
+"""
+function binompdf(n::Integer, p::AbstractFloat, x::Integer)
```
andreasnoack: Either this should be restricted to `p::Float64` or some more care is needed the computations below.

src/binomial.jl
```diff
@@ -0,0 +1,103 @@
...
+    # Using Saddle Point Algorithm by Catherine Loader which can be found here: http://octave.1599824.n4.nabble.com/attachment/3829107/0/loader2000Fast.pdf
+    if (n  < 1 || x > n || x < 0) return NaN end
+    if ( p < 0.0 || p > 1.0) return NaN end
+    if ( p == 0.0 ) return ( (x == 0) ? Float64(1.0):Float64(0.0)) end
```
andreasnoack: `Float64` not needed here since `1.0` is a `Float64` literal and please add spaces around `:`.

src/binomial.jl
```diff
@@ -0,0 +1,103 @@
...
+    if ( p == 1.0 ) return ( (x == n) ? Float64(1.0):Float64(0.0)) end
+    if ( x == 0 ) return exp(n*log(1-p)) end
```
andreasnoack: Use `n*log1p(-p)` here

src/binomial.jl
```diff
@@ -0,0 +1,103 @@
...
+    if ( x == n ) return exp(n*log(p)) end
+    n = convert(Int64, min(typemax(Int64), n))
```
andreasnoack: Is this silent truncation a good idea? Why not just throw if `n` is too large?

src/binomial.jl
```diff
@@ -0,0 +1,103 @@
...
+    p = convert(Float64, max(eps(Float64), p))
```
andreasnoack: Why truncate here?

src/binomial.jl
```diff
@@ -0,0 +1,103 @@
...
+    x = convert(Int64, min(typemax(Int64), x))
```
andreasnoack: Same as above

src/binomial.jl
```diff
@@ -0,0 +1,103 @@
...
+    lc = stirlerr(n) - stirlerr(x) - stirlerr(n-x) -D(x, n*p) - D(n-x, n*(1.0-p))
```
andreasnoack: Avoid floating point literal in `1.0 - p`. Just use `1 - p` and promotion will handle the rest.

src/binomial.jl
```diff
@@ -0,0 +1,103 @@
...
+    return  exp(lc)*sqrt(n/(2*pi*x*(n-x)))
+end
+
+function D(x::Int64, np::Float64) # Deviance term, x*log(x/np) + np - x
+    if( abs(x - np) < 0.1*(x+np))
+        s = (x - np)*(x - np)/(x + np)
+        v = (x - np)/(x + np)
+        ej = 2*x*v
+        j = 1
+        s1 = 0
```
andreasnoack: Should be `s1 = 0.0` since it will become a `Float64` in the loop.

src/binomial.jl
```diff
@@ -0,0 +1,103 @@
...
+        while true
+            ej *= v*v
+            s1 = s+ej/(2*j+1)
```
andreasnoack: spaces around `+`

src/binomial.jl
```diff
@@ -0,0 +1,103 @@
...
+    if ( x == n ) return exp(n*log(p)) end
+    n = convert(Int64, min(typemax(Int64), n))
+    p = convert(Float64, max(eps(Float64), p))
+    x = convert(Int64, min(typemax(Int64), x))
+    lc = stirlerr(n) - stirlerr(x) - stirlerr(n-x) -D(x, n*p) - D(n-x, n*(1.0-p))
+    return  exp(lc)*sqrt(n/(2*pi*x*(n-x)))
+end
+
+function D(x::Int64, np::Float64) # Deviance term, x*log(x/np) + np - x
+    if( abs(x - np) < 0.1*(x+np))
```
andreasnoack: spaces around `+`

src/binomial.jl
```diff
@@ -0,0 +1,103 @@
...
+julia> binomlogpdf(13, 0.58, 7)
+-1.5703423542721349
+```
+# Arguments
+- `n::Integer`: total number of trials.
+- `p::Float`: probability of success of each trial.
+- `x::Integer`: number of successful trials.
+...
+"""
+function binomlogpdf(n::Integer, p::AbstractFloat, x::Integer)
```
andreasnoack: Same comments as above

dmbates:
@Ellipse0934 I appreciate your creating this pull request and providing the reference to Catherine Loader's implementation note.

You mentioned in a comment that in #35 I provided implementations of `poispdf` and `poislogpdf` and you suggested using Loader's implementation instead.

The reason for having native Julia implementations that use commonly available functions like `lgamma`, `lbeta` is to allow for the function to be applied to more numeric types.  Two important types are `BigFloat` and dual numbers as implemented in `ForwardDiff.Dual`.  The `Dual` case is not important for discrete probability mass functions but the `BigFloat` case is.  If you need accuracy then you can switch to BigFloat at the expense of speed.

src/binomial.jl
```diff
@@ -0,0 +1,103 @@
+function stirlerr(n::Int64)
+    # stirlerr(x) = log(x!) -log(sqrt(2*pi*x)*(n/e)^n)
+    const S0 = 0.083333333333333333333        # 1/12
+    const S1 = 0.00277777777777777777778      # 1/360
+    const S2 = 0.00079365079365079365079365   # 1/1260
+    const S3 = 0.000595238095238095238095238  # 1/1680
+    const S4 = 0.0008417508417508417508417508 # 1/1188
+    # 35 values, sfe(x) = log(n!*e^n/((n^n)*sqrt(2*pi*n)))
+    sfe = (0.0810614667953272582196702, 0.0413406959554092940938221,  0.02767792568499833914878929,
```
dmbates: The comment should read '15', not '35'.  Also you use `sfe(x)` but the expression is in terms of `n`

A more Julian approach would be to create and store `sfe` as a `const` vector in the package, not in the function.  You could write it as
```julia
julia> const Stirlingerror = Float64[log(factorial(n)) - log(sqrt(2pi * n) * (n/Base.e)^n) for n in big(1):big(20)]
20-element Array{Float64,1}:
 0.0810615 
 0.0413407 
 0.0276779 
 0.0207907 
 0.0166447 
 0.0138761 
 0.0118967 
 0.0104113 
 0.00925546
 0.00833056
 0.00757368
 0.00694284
 0.00640899
 0.00595137
 0.00555473
 0.00520766
 0.0049014 
 0.00462915
 0.00438556
 0.00416632

julia> Stirlingerror[1]
0.08106146679532727
```
The calculation will be done in 256 bit precision then rounded to the 52 bits of precision for Float64 so it should be accurate.

This evaluation will be done at compile time.  If you change the condition for the explicit value branch in the code to
```julia
x < length(Stirlingerror) && return Stirlingerror[Int(x)]
```
then you can change the number of explicit values easily, in case you want to experiment with a cutoff.

Also, I would write the coefficients as `inv(12)` etc. if you want to use them as multiples.  It probably makes more sense to write the polynomial in terms of the inverse of the coefficients though.

dmbates: I made a mistake. The condition should be
```julia
x  ≤ length(Stirlingerror) && return Stirlingerror[Int(x)]
``

**Comment by dmbates** [comment]
Note that the `misc.jl` file already contains a `lstirling_asym` function for Float64 and Float32 types so the `sfe` function may be redundant.

**Comment by Ellipse0934** [comment]
@dmbates , You are correct about the stirlerr function. There is one issue, see that `lstriling_asym(1) = 0.23618..` and `stirlerr(1) = 0.08106..` due to this were failing. However it is definitely more accurate. So I guess its best to use the `sfe[]` lookup table for values under 16 and then shift to `lstirling_asym()`.

File: src/binomial.jl [hunk 15]
```diff
@@ -0,0 +1,98 @@
+#sfe(x) = log(n!*e^n/((n^n)*sqrt(2*pi*n))), for stirlerr
+const sfe = Float64[log(factorial(n)) - log(sqrt(2pi * n) * (n/Base.e)^n) for n in big(1):big(20)]
+# stirlerr(x) = log(x!) -log(sqrt(2*pi*x)*(n/e)^n)
+function stirlerr(n::Float64)
+    if (n <= 15)
+        @inbounds return(sfe[Int64(n)])
+    end
+    return lstirling_asym(n)
+end
+
```
dmbates: I think it should be `lstirling_asym(n + 1)`.  The `lstirling_asym` function returns the error from Stirling's approximation to `lgamma`, not `log(factorial(n)`

src/binomial.jl
```diff
@@ -0,0 +1,98 @@
...
+        error("n is too large.")
+    end
+    n = convert(Int64, n)
+    p = convert(Float64, p)
+    x = convert(Int64, x)
+    lc = stirlerr(n) - stirlerr(x) - stirlerr(n-x) -D(x, n*p) - D(n-x, n*(1 - p))
+    return  exp(lc)*sqrt(n/(2*pi*x*(n - x)))
+end
+
+function D(x::Int64, np::Float64) # Deviance term, x*log(x/np) + np - x
```
simonbyrne: maybe add a note as to how this is derived in the docs?

simonbyrne: with a link to the appropriate DLMF formula

src/binomial.jl
```diff
@@ -0,0 +1,98 @@
...
+...
+"""
+function binompdf(n::Integer, p::AbstractFloat, x::Integer)
+    # Using Saddle Point Algorithm by Catherine Loader which can be found here: http://octave.1599824.n4.nabble.com/attachment/3829107/0/loader2000Fast.pdf
+    if (n  < 1 || x > n || x < 0) return NaN end
+    if ( p < 0.0 || p > 1.0) return NaN end
+    if ( p == 0.0 ) return ( (x == 0) ? 1.0 : 0.0) end
+    if ( p == 1.0 ) return ( (x == n) ? 1.0 : 0.0) end
+    if ( x == 0 ) return exp(n*log1p(-p)) end
+    if ( x == n ) return exp(n*log(p)) end
```
simonbyrne: why not `p^n`?

src/binomial.jl
```diff
@@ -0,0 +1,98 @@
+#sfe(x) = log(n!*e^n/((n^n)*sqrt(2*pi*n))), for stirlerr
+const sfe = Float64[log(factorial(n)) - log(sqrt(2pi * n) * (n/Base.e)^n) for n in big(1):big(20)]
+# stirlerr(x) = log(x!) -log(sqrt(2*pi*x)*(n/e)^n)
```
simonbyrne: add this to the docs of the function, and link to the DLMF series

src/binomial.jl
simonbyrne: Certainly: more care with types is needed.

src/binomial.jl
Ellipse0934: I'm afraid there seems to be an error here. The `lstirling_asym()` is returning the error to the `ln(factorial(n))` instead of `ln(gamma(n))`. 
`f(x) = log(factorial(x)) - x*log(x) + x - 0.5*log(2*pi*x)`
`f(10) = 0.008330563433359472`
`lstirling_asym(10) = 0.00833056343336287`
`lstirling_asym(11) = 0.00757367548795184`

Ellipse0934:
Tests Failing currently because `lstirling_asym()` is not accurate for PR #36  to be accepted. Tests should pass then.

simonbyrne:
@andreasnoack can you kick the Appveyor CI?

andreasnoack:
> @andreasnoack can you kick the Appveyor CI?

Done

src/binomial.jl
```diff
@@ -0,0 +1,96 @@
...
+    x == 0 && return exp(n*log1p(-p))
+    x == n && return p^n
+    p = Float64(p)
+    (n, x) = (Int64(n), Int64(x))
+    lc = stirling_err(n) - stirling_err(x) - stirling_err(n - x) -D(x, n*p) - D(n - x, n*(1 - p))
+    return  T(exp(lc)*sqrt(n/(2π*x*(n - x))))
+end
+
+#
+const Stirlingerror = Float64[log(factorial(n)) - log(sqrt(2pi * n) * (n/Base.e)^n) for n in big(1):big(20)]
```
simonbyrne: can we switch this over the `stirling_asym` now?

src/binomial.jl
```diff
@@ -0,0 +1,96 @@
...
+```
+# Arguments
+- `n::Integer`: total number of trials.
+- `p::Float`: probability of success of each trial.
+- `x::Integer`: number of successful trials.
+...
+"""
+function binompdf{ T <: Union{Float16, Float32, Float64} }(n::Integer, p::T, x::Integer)
+    n > typemax(Int64) && error("n is too large.")
+    n < 1 && throw(ArgumentError("n = $n must be a positive integer"))
```
simonbyrne: should still be well-defined if `n == 0`.

src/binomial.jl
Ellipse0934: No. Still not accurate enough.

src/binomial.jl
simonbyrne: (and add a test-case)

src/binomial.jl
```diff
@@ -0,0 +1,96 @@
...
+    ( x > n || x < 0 ) && throw(ArgumentError("x = $x must ∈ [0, n]"))
+    (p < 0.0 || p > 1.0) && throw(ArgumentError("p = $p must ∈ [0, 1]"))
+    p == 0.0 && return ( (x == 0) ? T(1.0) : T(0.0) )
```
simonbyrne: use `one(T)`/`zero(T)`

src/binomial.jl
```diff
@@ -0,0 +1,96 @@
...
+    p == 1.0 && return ( (x == n) ? T(1.0) : T(0.0) )
+    x == 0 && return exp(n*log1p(-p))
+    x == n && return p^n
+    p = Float64(p)
+    (n, x) = (Int64(n), Int64(x))
```
simonbyrne: why do we need to convert this to `Int64`?

src/binomial.jl
```diff
@@ -0,0 +1,96 @@
...
+    return  T(exp(lc)*sqrt(n/(2π*x*(n - x))))
+end
+
+#
+const Stirlingerror = Float64[log(factorial(n)) - log(sqrt(2pi * n) * (n/Base.e)^n) for n in big(1):big(20)]
+function stirling_err(n)
+    n <= 20 && return Stirlingerror[n]
+    return lstirling_asym(n)
+end
+# Deviance term: D(x, np) = x*log(x/np) + np - x
```
simonbyrne: is it be more accurate to use `x*logmxp1(x/np)`?

src/binomial.jl
Ellipse0934: This is used by the function `D`, I am unsure of the precision implications it will have on the final output. We need to make more tests, investigate the precision and then make changes.

src/binomial.jl
simonbyrne: Ah, in that case it would be better to modify these lines:
https://github.com/JuliaStats/StatsFuns.jl/blob/5321025dd159585c870f784e800b9378b10a35ff/src/misc.jl#L60-L61
to fix it.

src/binomial.jl
Ellipse0934: I think you meant: `D(x, np) = x*logmxp1(x/np) -2x + x*x/np + np`. I roughly checked and that does not seem to be the case.

simonbyrne: Ah, no, it should be `-x*logmxp1(np/x) = -x*log(np/x)+x*(np/x)-x = x*log(x/np)+np-x`

src/binomial.jl
Ellipse0934: That was silly of me, no need to convert.

src/binomial.jl
Ellipse0934: But that will add discontinuity right ? Is that fine ?

simonbyrne: What discontinuity did you have in mind?

Ellipse0934: nevermind. I mixed up the versions on my computer. Will be replaced.

src/binomial.jl
Ellipse0934: Yes, this is giving better precision. Changed.

test/binomial.jl
```diff
@@ -0,0 +1,28 @@
...
+using Base.Test
+
+#Test Compares the difference in values w.r.t Rmath implementation
+
+N = Int64.( (0, 1, 1, 8, 20, 20, 20, 150, 700, 900, 1000) );
+P = (0.0, 0.5, 0.7, 0.6, 0.34, 0.89, 0.53, 0.77, 0.98, 0.5, 0.29);
+
+@testset "binompdf" begin
+for i = 1:length(N)
+    for x = Int64(0):N[i]
```
ron-wolf, line 11: Isn’t this inferred to be `Int64` by default?

Ellipse0934: Yep, 2 year old PR, back when I had no idea how Julia worked, will fix this tomorrow.

devmotion:
Since StatsFuns 0.9.11 (or more concretely https://github.com/JuliaStats/StatsFuns.jl/pull/125) a native Julia implementation is used for `binompdf` and `binomlogpdf`.

PR #35, opened 2018-03-06
Add a generic Poisson pdf and logpdf

dmbates:
Code much cleaner thanks to a suggestion by @andreasnoack


src/distrs/pois.jl
```diff
@@ -11,3 +11,12 @@ import .RFunctions:
     poisinvccdf,
     poisinvlogcdf,
     poisinvlogccdf
+
+# generic versions
+poispdf(λ::Real, x::Real) = exp(poislogpdf(λ, x))
+
+function poislogpdf(λ::T, x::T) where {T <: Real}
+    iszero(λ) ? (iszero(x) ? x : T(-Inf)) : x * log(λ) - λ - lgamma(x + 1)
```
andreasnoack: StatsFuns has the `xlogy` function exactly for this purpose

dmbates: Right.  In fact, I think I added it long, long ago.

dmbates:
CI failures on nightly are b/c the lbeta method for ForwardDiff.Dual uses a digamma which is not accessible.

Can I go ahead and merge?

src/distrs/pois.jl
```diff
@@ -11,3 +11,12 @@ import .RFunctions:
     poisinvccdf,
     poisinvlogcdf,
     poisinvlogccdf
+
+# generic versions
+poispdf(λ::Real, x::Real) = exp(poislogpdf(λ, x))
+
+function poislogpdf(λ::T, x::T) where {T <: Real}
+    iszero(λ) ? (iszero(x) ? x : T(-Inf)) : xlogy(x, λ) - λ - lgamma(x + 1)
```
andreasnoack: I don't think the ternaries are needed here when you are using `xlogy`.
```julia
poislogpdf(λ::T, x::T) where {T <: Real} = xlogy(x, λ) - λ - lgamma(x + 1)
```
should be sufficient.

dmbates: Right.  Thanks.

PR #36, opened 2018-03-08
Add doc strings and explicit BigInt method for lstirling_asym.

dmbates:
My purpose in creating an explicit `BigInt` method for `lstirling_asym` was to get full accuracy on small integer values of `x` from the `Float64` and `Float32` methods by evaluating `log(factorial(x-1))` explicitly and storing these in an array.  The only downside I can see to this is that the function will be a bit discontinuous at these small integer values (perhaps literally a bit or two of discontinuity).

This way the Stirling error function `stirlerr(n)` in #33 can be replaced by `lstirling_asym(n + 1)`

### DIFF
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -211,4 +211,4 @@ end
 Return the [`softmax transformation`](https://en.wikipedia.org/wiki/Softmax_function) applied to `x`
 """
 softmax!(x::AbstractArray{<:AbstractFloat}) = softmax!(x, x)
-softmax(x::AbstractArray{<:Real}) = softmax!(Array{Float64}(size(x)), x)
+softmax(x::AbstractArray{<:Real}) = softmax!(similar(x, Float64), x)
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -1,45 +1,81 @@
 # common facilities
 
 # scalar functions
+"""
+    xlogx(x::Real)
 
+Return `x * log(x)` for `x ≥ 0`.  (Special case is `x = 0`.)
+"""
 xlogx(x::Real) = x > zero(x) ? x * log(x) : zero(log(x))
 
+"""
+    xlogy(x::Real, y::Real)
+
+Return `x * log(y)` for `y > 0` with correct limit at `x = 0`.
+"""
 xlogy(x::T, y::T) where {T<:Real} = x > zero(T) ? x * log(y) : zero(log(x))
 xlogy(x::Real, y::Real) = xlogy(promote(x, y)...)
 
-# logistic: 1 / (1 + exp(-x))
-#
-logistic(x::Real) = one(x) / (one(x) + exp(-x))
+"""
+    logistic(x::Real)
 
-# logit: log(x / (1 - x))
-#
+Return `inv(exp(-x) + one(x))`, the logistic transformation of `x`
+"""
+logistic(x::Real) = inv(exp(-x) + one(x))
+
+"""
+    logit(x::Real)
+
+Return the log-odds, `log(x / (one(x) - x))`, for `0 < x < 1`
+"""
 logit(x::Real) = log(x / (one(x) - x))
 
-# log1psq: log(1+x^2)
-#
-log1psq(x::Real) = log1p(abs2(x))
-log1psq(x::Union{Float32,Float64}) = (ax = abs(x); ax < maxintfloat(x) ? log1p(abs2(ax)) : 2 * log(ax))
+"""
+    log1psq(x::Real)
 
-# log1pexp: log(1+exp(x))
-#
+Return `log(1+x^2)` evaluated carefully for abs(x) very small or very large
+"""
+log1psq(x::Real) = log1p(abs2(x))
+function log1psq(x::Union{Float32,Float64}) 
+    ax = abs(x)
+    ax < maxintfloat(x) ? log1p(abs2(ax)) : 2 * log(ax)
+end
+
+"""
+    log1pexp(x::Real)
+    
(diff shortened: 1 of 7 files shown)


src/basicfuns.jl
```diff
@@ -1,45 +1,81 @@
 # common facilities
 
 # scalar functions
+"""
+    xlogx(x::Real)
 
+Return `x * log(x)` for `x ≥ 0`.  (Special case is `x = 0`.)
```
simonbyrne: Maybe 

    Return `x * log(x)` for `x ≥ 0`, handling the limit as `x==0`.
    
    ```jldoctest
    julia> StatsFuns.xlogx(0)
    0.0
    ```

simonbyrne:
I've made a few minor tweaks.

src/basicfuns.jl
```diff
@@ -115,8 +173,11 @@ function _log1pmx_ker(x::Float64)
 end
 
 
-## logsumexp
+"""
+    logsumexp(x::Real, y::Real)
```
simonbyrne: we should probably rename this `logaddexp`?

src/misc.jl
```diff
@@ -43,7 +72,8 @@ function lstirling_asym(x::Float64)
 end
 
 function lstirling_asym(x::Float32)
-    t = 1f0/(x*x)
+    isinteger(x) && (0 < x ≤ length(lstirlingF32)) && return lstirlingF32[Int(x)]
+    t = inv/(abs2(x))
```
ararslan: Do you mean `inv(abs2(x))`?

src/distrs/pois.jl
```diff
@@ -18,3 +18,13 @@ poispdf(λ::Real, x::Real) = exp(poislogpdf(λ, x))
 poislogpdf(λ::T, x::T) where {T <: Real} = xlogy(x, λ) - λ - lgamma(x + 1)
 
 poislogpdf(λ::Number, x::Number) = poislogpdf(promote(float(λ), x)...)
+
+#=
+function poislogpdf(λ::Union{Float32,Float64}, x::Union{Float64,Float32,Integer})
```
andreasnoack: ❓

src/misc.jl
```diff
@@ -13,23 +14,51 @@ function logmvgamma(p::Int, a::Real)
...
+
+Relative error of approximation is bounded by
+    (174611/125400 x^-19) / (1/12 x^-1 - 1/360 x^-3)
+which is < 1/2 ulp for x >= 10.0
+total numeric error appears to be < 2 ulps
+"""
+function lstirling_asym end
+
+# For BigInt use the exact calculation of factorial(x-1)
+function lstirling_asym(x::BigInt)
```
andreasnoack: I think `lgamma` is fine everywhere
```julia
julia> norm([log(factorial(n)) - lgamma(n + 1) for n in big(1):big(500)])
0.0
```
Maybe better to make change the signature to `BigFloat` to match the other versions.

src/basicfuns.jl
dmbates: Look a bit further down.  There is a method for `x::AbstractArray{<:Real}` for which `sum` as part of the name is appropriate.

src/misc.jl
dmbates: Thanks for catching that.

dmbates:
I will add more tests as part of this PR.  I just noticed that the author of the `tvpack.jl` source file didn't include a single test.  If he had he would have discovered that there is a dependency on the `Distributions` package that is not satisfied.

dmbates:
I see now that `src/tvpack.jl` is not included by the file `src/StatsFuns.jl`.  Is `tvpack.jl` just there so that it doesn't get lost?  It does have the effect of lowering the coverage statistics considerably.

simonbyrne:
regarding the `lstirling_asym`, I'm not sure what we should do for smaller non-integer values (when the asymptotic error is too large).

The two most obvious choices seem to be the [Lanczos'](https://en.wikipedia.org/wiki/Lanczos_approximation) and [Spouge's](https://en.wikipedia.org/wiki/Spouge%27s_approximation) approximations. They are similar, but the former converges faster (and is used by Boost), whereas the latter can be computed to arbitrary precision (and is used by MPFR).

dmbates:
@simonbyrne My purpose in looking at `lstirling_asym` was because of Loader's technique of evaluating `binomlogpdf` as described in #33.  For `binomlogpdf` and for `poislogpdf` evaluation is always at positive integers and it is reasonable to store a vector of values up to the point where the `lstirling_asym` can be assumed to be accurate.

Do you know of other places where `lstirling_asym` is used?

simonbyrne:
At the moment, no. But if we can define it for real values we can use it for the gamma and beta densities as well.

simonbyrne:
I've tweaked the Stirling docs slightly, but lets merge this and we can flesh out other changes later.

src/misc.jl
```diff
@@ -17,34 +17,39 @@ end
...
-
+The remainder term after
 [Stirling's approximation](https://en.wikipedia.org/wiki/Stirling%27s_approximation)
-to the log-gamma (`lgamma`) function also applies to `log(factorial(k))`  through
-the relationship `log(factorial(k)) = lgamma(k + 1)`.  Stirling's approximation is
+to [`lgamma`](@ref):
 
-    lgamma(x) ≈ x*log(x) - x + log(2π/x)/2 = log(x)*(x-0.5) + log2π/2 -x
+```math
+\log \Gamma(x) \approx x \log(x) - x + log(2π/x)/2 = \log(x)*(x-1/2) + \log(2\pi)/2 - x
```
dmbates: Missed one `log` -> `\log`

PR #45, opened 2018-06-05
Write logsumexp in a more generic form

simonbyrne:
This uses Julia's reduction functionality to improve the `logsumexp`
functionality. It no longer relies on 1-based indexing, is fast
on CuArrays, and there is now a generic fallback for iterable collections.

I've also renamed the 2-arg `logsumexp` to `logaddexp`, since Julia
convention is to use different functions for reductions and
reducers (e.g. `max` vs `maximum`).

### DIFF
--- a/src/StatsFuns.jl
+++ b/src/StatsFuns.jl
@@ -44,7 +44,8 @@ export
     invsoftplus,    # alias of logexpm1
     log1pmx,        # log(1 + x) - x
     logmxp1,        # log(x) - x + 1
-    logsumexp,      # log(exp(x) + exp(y)) or log(sum(exp(x)))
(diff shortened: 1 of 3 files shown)


tpapp:
Very nice work! I think that `logaddexp` makes a lot of sense. Please consider adding tests for all combinations of `Inf`, `-Inf`, `NaN`. Not because I doubt that your code is correct, but to ensure that all corner cases are covered in case of future changes to this code.

simonbyrne:
They were pretty exhaustively tested already (I was able to remove some of the branches since we fixed `maximum` to be `NaN` absorbing). Are there any other cases you think need adding?

PR #48, opened 2018-06-26
Find lgamma in SpecialFunctions.jl for 0.7

### LINKED ISSUE #47: Base.lgamma
On Version 0.7.0-alpha.209, I do
```using SpecialFunctions, StatsFuns```

and get the message

```WARNING: importing deprecated binding Base.lgamma into StatsFuns``` followed by an error that points to the ```lstirling_asym``` function in misc.jl

roshanshariff:
Fixes #47

### DIFF
--- a/src/StatsFuns.jl
+++ b/src/StatsFuns.jl
@@ -4,6 +4,7 @@ module StatsFuns
 
 import Base: Math.@horner, @irrational
 using Compat
+using SpecialFunctions
 
 export
     # constants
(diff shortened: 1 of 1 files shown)


andreasnoack:
It looks like there are more issues with special function on master. Would you mind trying to fix them as well?

roshanshariff:
I think the test failure is coming from ForwardDiff.jl. I'll take a look there when I have time.

andreasnoack:
Thanks for looking more carefully.

PR #54, opened 2018-09-14
Promotions

bdeonovic:
These have been sitting around for awhile and I'd really like them pulled in.


devmotion:
Many/most of these changes were addressed in https://github.com/JuliaStats/StatsFuns.jl/pull/125 it seems.

Issue #52, opened 2018-07-04
log1pexp(x) for x < -37


cossio:
I was reading R's implementation of `log1pexp` and they include one additional branch (see https://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf, equation 10). 

Currently StatsFuns does this 
(https://github.com/JuliaStats/StatsFuns.jl/blob/ae40bc1c6ee63a5624ba6cd168da5bf43f473e59/src/basicfuns.jl#L66):

`log1pexp(x::Real) = x < 18.0 ? log1p(exp(x)) : x < 33.3 ? x + exp(-x) : oftype(exp(-x), x)`

Whereas including the additional branch it would be:

`log1pexp_v2(x::Real) = x < -37. ? exp(x) : x < 18. ? log1p(exp(x)) : x < 33.3 ? x + exp(-x) : x`

Apparently including the `x < -37.` check makes reduces the function's computation time by 25%. But I am no expert doing this kind of performance tests so please can someone else more knowledgeable check this? Here is the test I did:

```julia
using BenchmarkTools
@btime for x = -100. : 1. : 100., _ = 1 : 100
    log1pexp_v2(x)
end   #  310.471 μs (0 allocations: 0 bytes)
@btime for x = -100. : 1. : 100., _ = 1 : 100
    log1pexp(x)
end  #   423.671 μs (0 allocations: 0 bytes)
```

If this is worth it I can prepare a PR.

dmbates:
I can see how your version will be faster because it avoids the `exp(-x)` in `oftype(exp(-x), x)`.  You may consider adding
```julia
log1pexp_v2(x::Number) = log1pexp_v2(float(x))
```
to get coverage for integer arguments.

cossio:
I would do `log1pexp_v2(x::Real) = log1pexp_v2(float(x))`, since we don't want to cover complex arguments.

dmbates:
You're right.  That's better.

cossio:
Should I prepare a PR on this?

dmbates:
Yes, please.

cossio:
@dmbates Please see https://github.com/JuliaStats/StatsFuns.jl/pull/70

PR #57, opened 2018-10-09
log1pexp, #52

cossio:
https://github.com/JuliaStats/StatsFuns.jl/issues/52

### DIFF
--- a/test/basicfuns.jl
+++ b/test/basicfuns.jl
@@ -27,6 +27,11 @@ end
     @test log1pexp(-2.0)   ≈ log(1.0 + exp(-2.0))
     @test log1pexp(10000)  ≈ 10000.0
     @test log1pexp(-10000) ≈ 0.0
+
+    @test log1pexp(2f0)      ≈ log(1f0 + exp(2f0))
+    @test log1pexp(-2f0)     ≈ log(1f0 + exp(-2f0))
+    @test log1pexp(10000f0)  ≈ 10000f0
+    @test log1pexp(-10000f0) ≈ 0f0
 end
 
 @testset "log1mexp" begin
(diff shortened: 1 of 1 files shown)


cossio:
Merge this? @simonbyrne

src/basicfuns.jl
```diff
@@ -63,7 +63,7 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
 This is also called the ["softplus"](https://en.wikipedia.org/wiki/Rectifier_(neural_networks))
 transformation, being a smooth approximation to `max(0,x)`. Its inverse is [`logexpm1`](@ref). 
 """
-log1pexp(x::Real) = x < 18.0 ? log1p(exp(x)) : x < 33.3 ? x + exp(-x) : oftype(exp(-x), x)
+log1pexp(x::Real) = x ≤ -37. ? exp(x) : x ≤ 18. ? log1p(exp(x)) : x ≤ 33.3 ? x + exp(-x) : float(x)
```
tpapp: I wonder if it would be better to preserve the `oftype`, since in practice not all subtypes of `Real` implement `float` properly (in particular, some AD packages).
```suggestion
log1pexp(x::Real) = x ≤ -37. ? exp(x) : x ≤ 18. ? log1p(exp(x)) : x ≤ 33.3 ? x + exp(-x) : oftype(exp(-x), x)
```

simonbyrne: Actually, probably better to do something along the lines of:
```suggestion
log1pexp(x::Real) = log1pexp(float(x))
log1pexp(x::Float64) = x ≤ -37.0 ? exp(x) : 
                       x ≤ 18.0 ? log1p(exp(x)) :
                       x ≤ 33.3 ? x + exp(-x) :
                       x
```

cossio: > Actually, probably better to do something along the lines of:

But would that work with the AD packages @tpapp is referring to? It still depends on `float`.
As for the `oftype(exp(-x), x)` call, it has the downside that it performs an extra `exp` call which can be slow.

tpapp: I thought the compiler would elide `exp`, and just figure out the type, but you are right, it remains in there.

Perhaps `float` is OK, and the packages not supporting it should just be fixed.

cossio: So we agree that `float` is OK? Any issues with this or it can be merged?

cossio:
I'm sorry I had to close this in favor of #70 due to a slight git mess.

Issue #56, opened 2018-10-09
xexpx(z)


cossio:
I frequently encounter the function `xexpx(z) = z * exp(z)`, with the caveat that `xexpx(-Inf) == 0`. Thus `xexpx(z) = xlogx(exp(z))`, but compting it directly is faster because it avoids a superfluous `log` call.

Can we consider adding the definition:

```jula
xexpx(x::Real) = x == -Inf ? exp(x) : x * exp(x)
```

PR #58, opened 2018-10-09
add xexpx

cossio:
adds the functions xexpx and xexpy

https://github.com/JuliaStats/StatsFuns.jl/issues/56


cossio:
@simonbyrne Can anyone take a look here?

PR #62, opened 2018-10-26
Support for zero σ normals

simonbyrne:
This adds support for normal distributions with zero standard deviation. At the moment, this gives the same values as R. The only one I'm not particularly sure about is what to return for the quantile at 1 (R gives +Inf, there is a reasonable argument I think it should be μ).

### DIFF
--- a/src/distrs/norm.jl
+++ b/src/distrs/norm.jl
@@ -1,35 +1,92 @@
 # functions related to normal distribution
 
-xval(μ::Real, σ::Real, z::Number) = μ + σ * z
+function xval(μ::Real, σ::Real, z::Number)
+    if isinf(z) && iszero(σ)
+        μ + one(σ) * z
+    else
+        μ + σ * z
+    end
+end
 zval(μ::Real, σ::Real, x::Number) = (x - μ) / σ
 
 # pdf
 normpdf(z::Number) = exp(-abs2(z)/2) * invsqrt2π
-normpdf(μ::Real, σ::Real, x::Number) = normpdf(zval(μ, σ, x)) / σ
+function normpdf(μ::Real, σ::Real, x::Number)
+    if iszero(σ)
+        if x == μ
+            z = zval(μ, one(σ), x)
+        else
+            z = zval(μ, σ, x)
+            σ = one(σ)
+        end
+    else
+        z = zval(μ, σ, x)
+    end
+    normpdf(z) / σ
+end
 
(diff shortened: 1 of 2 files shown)


ararslan:
> there is a reasonable argument I think it should be μ

Can you elaborate? I would have thought `Inf` would be the most natural choice.

simonbyrne:
Because if σ is zero, the distribution is a point-mass at μ, and this is how we define quantiles in other cases.

That said, I'm not particularly bothered by it (there are also inconsistencies in how we handle the quantile at 0).

ararslan:
Ah, that does indeed make sense. I'd say go with what your heart tells you.

simonbyrne:
The easiest option is to go with what R does. We can always revisit this later.

andreasnoack:
The code complication reflects well that the distribution at `σ=0` is fundamentally different from when `σ>0`. We discussed this a while ago in the context of the Poisson where the consensus, at that time, was to exclude degenerate continuous distributions. Is hitting `σ=0` a common situation in practice? If it is then I'm okay with the change but the code complication is a bit unfortunate.

simonbyrne:
Yes, I'm seeing it in a case where `σ` underflows. Currently there are ugly hacks, like adding `eps()`, but this should make the code much clearer.

andreasnoack:
Okay. It is better to have the special cases here than in user code.

tpapp:
I have been dealing with a lot of similar cases lately in the context of Bayesian inference for a rather complex structural model in economics. Eg for an MCMC sampler that works on unconstrained ℝⁿ, the a variance (which is >0) would be `exp` of some coordinate in ℝ using standard transformations. Underflow is very easy to get during MCMC adaptation.

I tried to make all corner cases work like this PR, but eventually abandonned this and came to the conclusion that the best solution is to error since there is no clearly good choice. Eg if `σ` is a `ForwardDiff.Dual` or similar, I am not sure that the derivatives will make sense (currently one can AD through most distributions for finite, valid values). The user should check for these things and do the appropriate thing (eg reject the parameters for MCMC). It is best not to catch the error to call attention to this.

(PS: realize this was merged while typing this, sorry for the noise).

simonbyrne:
@tpapp Care to share some examples? Handling of invalid values has been one of the biggest challenges with the "distributions as objects" approach, so I'd be keen to learn more about how problems arise.

PR #66, opened 2019-01-25
Provides same interface to studentized range distribution functions ([p/q]tukey)

BioTurboNick:
To support an implementation of proposed `StudentizedRange()` in Distributions.jl.

The interface of the Rmath library is a bit different for the tukey functions than the others, with a parameter (nranges) set to 1 which isn't really explained by the R documentation. Thanks to the Rmath.jl package which showed me the right argument ordering.

All are a bit slower than the other distributions, ~1 order of magnitude.

Note: there is a case where the underlying ptukey and qtukey functions diverge with higher df/k and values closer to 0. This is documented as a broken test and with a special case in testing to avoid those values.


src/distrs/srdist.jl
```diff
@@ -0,0 +1,30 @@
...
+    srdistlogcdf,
+    srdistlogccdf,
+    srdistinvcdf,
+    srdistinvccdf,
+    srdistinvlogcdf,
+    srdistinvlogccdf
+
+# pdf for numbers with generic types
+function srdistpdf(ν::Real, k::Real, x::Number)
+    Φ(x) = (1+erf(x / √2)) / 2
```
simonbyrne: isn't this just `normcdf`?

src/distrs/srdist.jl
```diff
@@ -0,0 +1,30 @@
...
+    ϕ(x) = derivative(Φ, x)
```
simonbyrne: `normpdf`?

simonbyrne: No, I meant that if `Φ` (upper case) is the normal distribution CDF, then `ϕ` (lower case) is the normal distribution pdf.

BioTurboNick: Ah yes, you're right.

simonbyrne: You can also get rid of the ForwardDiff.jl dependency

src/distrs/srdist.jl
```diff
@@ -0,0 +1,29 @@
...
+    Φ(x) = normcdf(x)
+    ϕ(x) = normpdf(x)
+    function outer(y)
+        inner(u) = ϕ(u) * ϕ(u - x*y) * (Φ(u) - Φ(u - x*y))^(k-2)
+        inner_part = quadgk(inner, -Inf, Inf) |> first
+        return inner_part * y^ν * ϕ(y*√ν)
+    end
+    integral = quadgk(outer, 0.0, Inf) |> first
+    return integral * (√(2π) * k * (k-1) * ν^(ν/2)) / (gamma(ν/2) * 2^(ν/2 - 1))
+end
```
simonbyrne: I'm fine with adding the `p/q` functions, but computing the density via nested quadrature seems like a very poor way to go about this. I'd probably prefer to leave this part off for now.

BioTurboNick: Sure thing. It's not necessary for the stats tests.

PR #67, opened 2019-01-29
include tvpack.jl with bivariate normal cdf

magerton:
This reincludes the `tvpack.jl` file with the bivariate normal CDF function. I've added a `bvncdf(x,y,r)` function that corresponds to the standard normal cdf `Φ(x,y;ρ)` where `r=ρ`. The function also handles `Inf` and `-Inf`  arguments correctly now.

### DIFF
--- a/src/tvpack.jl
+++ b/src/tvpack.jl
@@ -1,3 +1,5 @@
+export bvncdf
+
 # Written in Julia by Andreas Noack Jensen
 # January 2013
 #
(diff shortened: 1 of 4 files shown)


src/tvpack.jl
```diff
@@ -1,3 +1,5 @@
+export bvncdf
+
```
devmotion, line 2: It's already exported in src/StatsFuns.jl it seems?

```suggestion
```

spaette:
Possibly `throw` could be misspelled

```
$ tail -n 15 StatsFuns.jl/src/tvpack.jl
    # here we either have 0 or the marginal
    else
        if dh == typemax(T) || dk == typemax(T)
            return zero(T)
        elseif dh == typemin(T)
            return normcdf(-dk)
        elseif dk == typemin(T)
            return normcdf(-dh)
        else
            thow(error())
        end
    end
end

bvncdf(dh, dk, r) = bvnuppercdf(-dh, -dk, r)
$

magerton:
> Possibly `throw` could be misspelled
> 
> ```
> $ tail -n 15 StatsFuns.jl/src/tvpack.jl
>     # here we either have 0 or the marginal
>     else
>         if dh == typemax(T) || dk == typemax(T)
>             return zero(T)
>         elseif dh == typemin(T)
>             return normcdf(-dk)
>         elseif dk == typemin(T)
>             return normcdf(-dh)
>         else
>             thow(error())
>         end
>     end
> end
> 
> bvncdf(dh, dk, r) = bvnuppercdf(-dh, -dk, r)
> $ 
> ```

Good catch, @spaette. Just submitted PR #160

PR #71, opened 2019-07-01
add some tests for `logmvgamma`

johnczito:
I guess I don't understand how this decreased coverage.

Anyway, let me know what I need to change.


andreasnoack:
Coverage can be a bit noisy. Thanks for adding these.

PR #74, opened 2019-09-13
Implement `logsumexp` with `dims` keyword arguments

xukai92:


src/basicfuns.jl
```diff
@@ -190,14 +190,21 @@ function logsumexp(X)
     isempty(X) && return log(sum(X))
     reduce(logaddexp, X)
 end
-function logsumexp(X::AbstractArray{T}) where {T<:Real}
+function logsumexp(X::AbstractVector{T}) where {T<:Real}
```
nalimilan: This function is now redundant with the new one.

xukai92: I don't have the corresponding line for `isfinite(u) || return float(u)` in the vector one. Any idea how to implement it?

xukai92:
> Thanks. I guess that can be useful, but could you check whether similar functions should accept the same argument? It would be a bit weird to only add it to this one.

By searching `sum` in the repo, `logsumexp` is the only thing turning up.

src/basicfuns.jl
xukai92: Maybe `u isa AbstractArray || isfinite(u) || return float(u)`?

nalimilan: Yes, sounds good. Basically, this is just an optimization to avoid computing the value when we already know it will be `NaN` or `Inf`, right?

xukai92: I believe so. I will unify the two by doing what I proposed.

xukai92: Done.

src/basicfuns.jl
```diff
@@ -190,12 +190,12 @@ function logsumexp(X)
...
-function logsumexp(X::AbstractArray{T}) where {T<:Real}
+function logsumexp(X::AbstractArray{T}; dims=:) where {T<:Real}
     isempty(X) && return log(zero(T))
-    u = maximum(X)
-    isfinite(u) || return float(u)
+    u = maximum(X; dims=dims)
+    u isa AbstractArray || isfinite(u) || return float(u)
     let u=u # avoid https://github.com/JuliaLang/julia/issues/15276
-        u + log(sum(x -> exp(x-u), X))
+        u .+ log.(sum(exp.(X .- u); dims=dims))
```
nalimilan: This is going to allocate a temporary array. Unfortunately, there's no way currently to compute the sum over dimensions of a lazy `Broadcasted` object: we need https://github.com/JuliaLang/julia/pull/31020 for this. Until then, better keep a branch with the existing code when `u` is an array, and leave a TODO.

src/basicfuns.jl
```diff
@@ -190,12 +190,12 @@ function logsumexp(X)
     isempty(X) && return log(sum(X))
     reduce(logaddexp, X)
 end
-function logsumexp(X::AbstractArray{T}) where {T<:Real}
+function logsumexp(X::AbstractArray{T}; dims=:) where {T<:Real}
     isempty(X) && return log(zero(T))
```
nalimilan: This won't be correct when `dims != :`, right? At this point, probably better add an internal `_logsumexp(X, dims)` function dispatching on `dims::Colon` vs. `dims::Any`.

src/basicfuns.jl
xukai92: By "a branch" you mean an `if` statement or something?

src/basicfuns.jl
nalimilan: Something like this:
```julia
logsumexp(X::AbstractArray{<:Real}; dims=:) = _logsumexp(X, dims)
_logsumexp(X::AbstractArray{<:Real}, dims::Colon) = ...
_logsumexp(X::AbstractArray{<:Real}, dims::Any) = ...
```

That's the pattern used in Base for most reduction functions.

xukai92: Actually the implementation works for `dims != :`. Things like `logsumexp(x; dims=[1,2])` also work. Do I still need this dispatch pattern here?

nalimilan: Well, returning `log(zero(T))` clearly isn't correct in that case...

xukai92: Oh I see what you are referring to! I misunderstood your point. 
Do we want to only do `isempty(X) && return log(zero(T))` for `dims::Colon`?

nalimilan: You need to handle that case also for other `dims`, since `maximum` throws an error for empty arrays. Or maybe a simpler solution is to use `reduce(max, X, dims=dims, init=log(zero(T)))` instead.

xukai92: I tried your solution.
```julia
    isempty(X) && return reduce(max, X, dims=dims, init=log(zero(T)))
    u = maximum(X; dims=dims)
    u isa AbstractArray || isfinite(u) || return float(u)
    let u=u # avoid https://github.com/JuliaLang/julia/issues/15276
        u .+ log.(sum(exp.(X .- u); dims=dims))
    end
end
```
gives
```julia
julia> X = zeros(0, 0)
0×0 Array{Float64,2}

julia> logsumexp(X)
-Inf

julia> logsumexp(X; dims=1)
1×0 Array{Float64,2}

julia> logsumexp(X; dims=[1,2])
1×1 Array{Float64,2}:
 -Inf
```

Is this something we want?

nalimilan: I guess, yes. That sounds consistent with what `sum` does.

But my suggestion was to use `reduce(max, X, dims=dims, init=log(zero(T)))` instead of `maximum`. It wouldn't make sense to use it only when `X` is empty.

xukai92: You mean replace both them empty check line and the `u = maximum...` with `reduce` right?

nalimilan: Right. Can you add tests with `dims=1`, `dims=2` and `dims=[1, 2]` when the output isn't finite? Otherwise that case isn't covered.

xukai92: Test for `dims=1` was there. I just added tests for `dims=2` and `dims=[1,2]` as well.

src/basicfuns.jl
nalimilan: This point is still relevant: please keep the current code when `u` isn't an array, as it's more efficient. Also leave a TODO.

xukai92:
Thanks for the help in this PR!

nalimilan:
I missed that the docstring should mention the new keyword argument. Could you make a PR to fix that?

PR #76, opened 2019-11-04
Fix deprecation warning

nilshg:
From report on Slack
```
┌ Warning: `lgamma(x::Real)` is deprecated, use `(logabsgamma(x))[1]` instead.
│   caller = lstirling_asym(::BigFloat) at misc.jl:56
└ @ StatsFuns ~/.julia/packages/StatsFuns/2QE7p/src/misc.jl:56
```

### DIFF
--- a/src/misc.jl
+++ b/src/misc.jl
@@ -53,7 +53,7 @@ which is < 1/2 ulp for x >= 10.0, and total numeric error appears to be < 2 ulps
 """
 function lstirling_asym end
 
-lstirling_asym(x::BigFloat) = lgamma(x) + x - log(x)*(x - big(0.5)) - log2π/big(2)
+lstirling_asym(x::BigFloat) = (logabsgamma(x))[1] + x - log(x)*(x - big(0.5)) - log2π/big(2)
 
 lstirling_asym(x::Integer) = lstirling_asym(float(x))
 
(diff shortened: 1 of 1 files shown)


src/misc.jl
```diff
@@ -53,7 +53,7 @@ which is < 1/2 ulp for x >= 10.0, and total numeric error appears to be < 2 ulps
 """
 function lstirling_asym end
 
-lstirling_asym(x::BigFloat) = lgamma(x) + x - log(x)*(x - big(0.5)) - log2π/big(2)
+lstirling_asym(x::BigFloat) = (logabsgamma(x))[1] + x - log(x)*(x - big(0.5)) - log2π/big(2)
```
nalimilan:
```suggestion
lstirling_asym(x::BigFloat) = logabsgamma(x)[1] + x - log(x)*(x - big(0.5)) - log2π/big(2)
```

nilshg:
Hm, trying to generate the Project.toml I get an error as the REQUIRE has Rmath 0.4.0, but the package manager will only allow me to install 0.5.0 - is that okay? Never had to port a REQUIRE before so not sure what's innocuous here and what isn't!

nalimilan:
Hmm, you shouldn't get an error if the minimum version is lower than the installed version, should you? :-/ Anyway, I guess it's OK to require 0.5.0 now.

andreasnoack:
Superseded by #77

PR #77, opened 2019-11-07
Upgrade to Project file and to SpecialFunctions 0.8

andreasnoack:
Supersedes #76. Regarding the changes to the `logmvgamma` tests, the old version didn't compute what the docstring claimed. Fixing the issue required changing the tests.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,0 +1,19 @@
+name = "StatsFuns"
+uuid = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
+version = "0.9"
+
+[deps]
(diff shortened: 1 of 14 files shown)


Project.toml
```diff
@@ -0,0 +1,19 @@
+name = "StatsFuns"
+uuid = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
+version = "0.9"
```
ararslan:
```suggestion
version = "0.9.0"
```
?

andreasnoack: Isn't that effectively the same thing?

ararslan: Yeah I suppose so, I just don't usually see packages declare their own versions that way.

PR #82, opened 2019-12-09
Fixes initial value for reduce in logsumexp

itsdfish:
This PR request fixes the initial value for `reduce` in `logsumexp`. This error was discovered in the following [issue](https://github.com/TuringLang/Turing.jl/issues/1017#issuecomment-563128527) in Turing (see final comment).

### DIFF
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -50,18 +50,18 @@ logit(x::Real) = log(x / (one(x) - x))
 Return `log(1+x^2)` evaluated carefully for `abs(x)` very small or very large.
 """
 log1psq(x::Real) = log1p(abs2(x))
-function log1psq(x::Union{Float32,Float64}) 
+function log1psq(x::Union{Float32,Float64})
     ax = abs(x)
     ax < maxintfloat(x) ? log1p(abs2(ax)) : 2 * log(ax)
 end
 
 """
     log1pexp(x::Real)
-    
+
 Return `log(1+exp(x))` evaluated carefully for largish `x`.
 
 This is also called the ["softplus"](https://en.wikipedia.org/wiki/Rectifier_(neural_networks))
-transformation, being a smooth approximation to `max(0,x)`. Its inverse is [`logexpm1`](@ref). 
+transformation, being a smooth approximation to `max(0,x)`. Its inverse is [`logexpm1`](@ref).
 """
 log1pexp(x::Real) = x < 18.0 ? log1p(exp(x)) : x < 33.3 ? x + exp(-x) : oftype(exp(-x), x)
 log1pexp(x::Float32) = x < 9.0f0 ? log1p(exp(x)) : x < 16.0f0 ? x + exp(-x) : oftype(exp(-x), x)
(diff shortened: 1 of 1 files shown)


src/basicfuns.jl
```diff
@@ -191,7 +191,7 @@ function logsumexp(X)
     reduce(logaddexp, X)
 end
 function logsumexp(X::AbstractArray{T}; dims=:) where {T<:Real}
-    u = reduce(max, X, dims=dims, init=log(zero(T)))
+    u = reduce(max, X, dims=dims, init=zero(T))
```
mohamed82008: This is not right, `X` can be all negative.

itsdfish: Thanks for catching the error. Do you know what the correct initial value should be?

mohamed82008: I think we should not have `init` at all. Just check if `X` is empty and return `log(zero(T))` then. Otherwise, keep going with the `reduce`.

itsdfish: Good call. I made the changes you suggested and the results hold for both negative and positive values. Please review if you can.

mohamed82008:
Would be nice to add a `ForwardDiff` test to catch such error if it happens again in the future.

nalimilan:
This isn't correct when `dims != :`. Can you describe the problem you're trying to fix (not knowing anything about what ForwardDiff is trying to do in that issue)? Anyway we'll need to add a test for it.

mohamed82008:
The problem is that `log(zero(ForwardDiff.Dual{Nothing}))` is `ForwardDiff.Dual{Nothing}(-Inf, NaN)`. The `NaN` then propagates to `u` when doing the `reduce`.

mohamed82008:
`ForwardDiff.gradient(logsumexp, rand(3))` now gives a vector of `NaN`.

nalimilan:
I guess we could use another way to get `-Inf` of the right type, e.g. `convert(T, -Inf)`. Would that work? Or is there another solution which is OK for ForwardDiff?

mohamed82008:
Yup `init=convert(T, -Inf)` works like a charm!

mohamed82008:
I am curious though, why is the solution here not correct when `dims != :`? Could you please give an example?

nalimilan:
> Yup `init=convert(T, -Inf)` works like a charm!

Cool. Actually this won't work for integer inputs, so better use something like `oftype(log(zero(T)), -Inf)`. Would you update the PR to use that? It would also be nice to have a test or at least a comment explaining why we don't use the simplest syntax.


> I am curious though, why is the solution here not correct when `dims != :`? Could you please give an example?

```julia
julia> reduce(max, zeros(0, 2), dims=1, init=log(zero(Int)))
1×2 Array{Float64,2}:
 -Inf  -Inf

julia> reduce(max, zeros(0, 2), dims=1)
ERROR: ArgumentError: reducing over an empty collection is not allowed

```

mohamed82008:
@nalimilan the `reduce` won't be called if `isempty(X)` is true  which it is in case of `zeros(0, 2)`.

mohamed82008:
I don't have write access to this PR, so @itsdfish please use:
```julia
u = reduce(max, X, dims=dims, init=oftype(log(zero(T)), -Inf))
```
instead of
```julia
u = reduce(max, X, dims=dims, init=log(zero(T)))
```
and remove all the other changes in this PR.

itsdfish:
The updates seem to work on my end.

nalimilan:
Thanks. One last check: is it really expected/correct that `log(ForwardDiff.Dual{Nothing}(0.0, 0.0))` gives `ForwardDiff.Dual{Nothing}(-Inf, NaN)`? Where does that NaN come from? Maybe we should discuss this with ForwardDiff developers?

Anyway, if we make this change, we should add a comment to ensure somebody doesn't revert this without knowing the reason why it exists.

itsdfish:
To make things easier, I added you both as collaborates in case you want to make changes.

mohamed82008:
Perhaps merge this?

nalimilan:
https://github.com/JuliaStats/StatsFuns.jl/pull/83

PR #87, opened 2019-12-26
Fix logarithm bugs

ericproffitt:
`binomlogpdf` was unable to handle `0log(0)`, fixed using the `xlogy` function.

using `log1p` with `logbeta` results in slight numerical inaccuracies, so changed to using all `loggamma`.

fixed `xlogy`, should return `zero(log(y))`.


nalimilan:
Thanks. Can you add tests for things which were previously incorrect?

ericproffitt:
Ok, so after running some tests, the situation was nastier than I thought.

I think I've got everything correct now.

It turned out the simplest way to do it was with a ternary operator which simply returns `convert(typeof(float(p)), -Inf)` if `k` isn't an integer between zero and `n`, inclusive. It also should be type stable, since the return type is always equal to `float(p)`.

Tests have now been included.

test/basicfuns.jl
```diff
@@ -6,6 +6,11 @@ using StatsFuns, Test
 
     @test iszero(xlogy(0, 1))
     @test xlogy(2, 3) ≈ 2.0 * log(3.0)
+
+    @test xlogy(-2, Base.MathConstants.e) ≈ -2
+    @test xlogy(0, 1) ≈ 0
+    @test xlogy(-1, 1) ≈ 0
+    @test xlogy(0, 0) ≈ 0
```
nalimilan: This isn't supposed to work according to the docstring. We could update it to say `y ≥ 0`, but then it wouldn't be right about the "correct limit" AFAICT. Maybe we should just drop that restriction and just say that zero is returned when `x == 0`. That's what [SciPy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.xlogy.html) does.

test/misc.jl
```diff
@@ -55,3 +55,26 @@ end
...
+        for p in push!(rand(50), [0, 1]...)
+            for n in 0:50
+                for k in push!(collect(Any, -50:50), randn(10)...)
+                    @test binomlogpdf(n, Float32(p), k) ≈ log(pdf(Binomial(n, p), k))
+                end
+            end
+        end
+    end
+end
+
```
nalimilan: Remove empty lines.

test/misc.jl
```diff
@@ -55,3 +55,26 @@ end
...
     end
 end
+
+@testset "binomlogpdf" begin
+    @testset "identity" begin
+
+        for p in push!(rand(50), [0, 1]...)
+            for n in 0:50
+                for k in push!(collect(Any, -50:50), randn(10)...)
+                    @test binomlogpdf(n, Float32(p), k) ≈ log(pdf(Binomial(n, p), k))
```
nalimilan: Isn't this a tautological test, since `Binomial` uses `binomlogpdf` under the hood?

src/distrs/binom.jl
```diff
@@ -16,4 +16,4 @@ import .RFunctions:
 binompdf(n::Real, p::Real, k::Real) = exp(binomlogpdf(n, p, k))
 
 # logpdf for numbers with generic types
-binomlogpdf(n::Real, p::Real, k::Real) = -log1p(n) - logbeta(n - k + 1, k + 1) + k * log(p) + (n - k) * log1p(-p)
+binomlogpdf(n::Real, p::Real, k::Real) = (isinteger(k) & (zero(k) <= k <= n)) ? convert(typeof(float(p)), loggamma(n + 1) - loggamma(k + 1) - loggamma(n - k + 1) + xlogy(k, p) + xlogy(n - k, 1 - p)) : convert(typeof(float(p)), -Inf)
```
nalimilan: Break this into multiple lines with a `if`. It would also be simpler/clearer to only apply the conversion once to the result, instead of repeating it.

In terms of performance, can you check the difference with the current implementation? I'm afraid it's significantly more complex, with the three calls to `loggamma`.

src/basicfuns.jl
```diff
@@ -18,7 +18,7 @@ xlogx(x::Real) = x > zero(x) ? x * log(x) : zero(log(x))
 
 Return `x * log(y)` for `y > 0` with correct limit at `x = 0`.
 """
-xlogy(x::T, y::T) where {T<:Real} = x > zero(T) ? x * log(y) : zero(log(x))
+xlogy(x::T, y::T) where {T<:Real} = x != zero(T) ? x * log(y) : zero(log(y))
```
nalimilan: BTW, maybe you could apply the same change to `xlogx`. That would be more consistent, and more in line with the docstring, which doesn't say that zero is returned for negative inputs at all.

test/basicfuns.jl
ericproffitt: I think the docstring is incorrect in this case. Maybe I'm misunderstanding the purpose of having a function like `xlogy`, but being able to evaluate 0log(0) is I believe intended to be its principal benefit.

I'm not sure I understand what you mean when you say,

> ...but then it wouldn't be right about the "correct limit" AFAICT

could you expand on this?

I dont believe dropping the constraint on `y` entirely is correct, as it should error out when `y` is of type negative real, irrespective of what `x` is.

My appraisal of the situation is that the docstring should be changed to `y≥0`.

test/misc.jl
ericproffitt: So `Distributions.binompdf` has two methods, if `p` is of type `Union{Int64, Float64}`, then it makes a `ccall` to an underlying C library, which, as far as I can tell, is bug free and always returns the correct answer; please see

_src/rmath.jl:70_

However when `p` is of type `Real` but not of type `Union{Int64, Float64}`, then it dispatches on the `binompdf`, found in

`src/distrs/binom.jl:16`

So the above testing code is simply checking that the methods match, under assumption that the method which dispatches on the C code is correct (which I'm pretty sure it is).

src/distrs/binom.jl
ericproffitt: > Break this into multiple lines with a if. It would also be simpler/clearer to only apply the conversion once to the result, instead of repeating it.

Will do!

As for performance, the current implementation of `logbeta` is

`logbeta(a::Number, b::Number) = loggamma(a)+loggamma(b)-loggamma(a+b)`

So if anything, the original implementation is actually less performant than the new version.

src/basicfuns.jl
ericproffitt: Strictly speaking I'm not sure this is necessary.

I'll do it if you really want to, but for `xlogx`, all changing the condition statement from `x > 0` to `x != 0` will do is make it so when `x` is negative the condition returns `true` and the negative log argument error occurs at `x * log(x)` rather than at `zero(log(x))`.

test/basicfuns.jl
```diff
@@ -6,6 +6,11 @@ using StatsFuns, Test
 
     @test iszero(xlogy(0, 1))
     @test xlogy(2, 3) ≈ 2.0 * log(3.0)
+
+    @test xlogy(-2, Base.MathConstants.e) ≈ -2
+    @test xlogy(0, 1) ≈ 0
```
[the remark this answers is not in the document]
ericproffitt: ok I used `==` when the function evaluates to zero.

test/misc.jl
ericproffitt: removed

src/distrs/binom.jl
ericproffitt: Ok made requested changes.

test/basicfuns.jl
ericproffitt: changed docstring to `y≥0`.

nalimilan: I meant that I don't think it's correct to say that the limit of `x * log(y)` when `x` tends towards zero is zero when `y == 0`, is it? `x * log(0)` isn't defined when `x > 0`, so it's not appropriate to speak about its limit at `0`.

test/misc.jl
nalimilan: OK, thanks. Though can you call `binomlogpdf` directly instead of using `log(pdf(Binomial(...), ...))`? `Binomial` is not defined in this package, but in Distributions. Maybe also add a comment explaining that you test the generic Julia definition against Rmath.

src/distrs/binom.jl
```diff
@@ -16,4 +16,12 @@ import .RFunctions:
 binompdf(n::Real, p::Real, k::Real) = exp(binomlogpdf(n, p, k))
 
 # logpdf for numbers with generic types
-binomlogpdf(n::Real, p::Real, k::Real) = -log1p(n) - logbeta(n - k + 1, k + 1) + k * log(p) + (n - k) * log1p(-p)
+function binomlogpdf(n::Real, p::Real, k::Real)
+    if isinteger(k) & (zero(k) <= k <= n)
+        x = loggamma(n + 1) - loggamma(k + 1) - loggamma(n - k + 1) + xlogy(k, p) + xlogy(n - k, 1 - p)
```
nalimilan: Break lines at 92 chars:
```suggestion
        x = loggamma(n + 1) - loggamma(k + 1) - loggamma(n - k + 1) +
            xlogy(k, p) + xlogy(n - k, 1 - p)
```

test/misc.jl
```diff
@@ -55,3 +55,16 @@ end
         end
     end
 end
+
+@testset "binomlogpdf" begin
+    @testset "identity" begin
+
+        for p in push!(rand(50), [0, 1]...)
```
nalimilan:
```suggestion
        for p in [rand(50); 0; 1]
```

test/misc.jl
```diff
@@ -55,3 +55,16 @@ end
...
+            for n in 0:50
+                for k in push!(collect(Any, -50:50), randn(10)...)
```
nalimilan:
```suggestion
                for k in Any[-50:50; randn(10)]
```

src/basicfuns.jl
nalimilan: Ah, right, I got confused. But for consistency I'd prefer to also change `xlogx`. Otherwise when reading the code people will wonder why there's a difference with `xlogy`.

test/basicfuns.jl
ericproffitt: Ok I see what you're getting at now, thanks for clarifying.

So strictly speaking you're right, the limit would not be defined. However I think the practical benefits of having `xlogy` evaluate 0log(0) as if you were actually taking the limit xlog(x) as x->0⁺ outweigh the slight lapse in mathematical rigor.

The docstring, as it currently stands, is:

`Return `x * log(y)` for `y > 0` with correct limit at `x = 0`.`

The mention of the limit at `x=0` being entirely superfluous when `y>0`, since it's unnecessary to consider any limits in this context, exact evaluation is always well-defined.

And to reiterate, if it isn't the problem of 0log(0) which we're trying to resolve, then it's not clear to me that the function `xlogy` provides any benefit over just simply evaluating `x * log(y)`.

nalimilan: Yes, it's not a big deal. Though we could also use a more explicit mention about what happens when `x == 0`.

src/basicfuns.jl
ericproffitt: change made.

test/misc.jl
ericproffitt: fixed.

test/basicfuns.jl
ericproffitt: ok, added a more descriptive docstring.

src/distrs/binom.jl
```diff
@@ -16,4 +16,13 @@ import .RFunctions:
...
-binomlogpdf(n::Real, p::Real, k::Real) = -log1p(n) - logbeta(n - k + 1, k + 1) + k * log(p) + (n - k) * log1p(-p)
+function binomlogpdf(n::Real, p::Real, k::Real)
+    if isinteger(k) & (zero(k) <= k <= n)
+        x = loggamma(n + 1) - loggamma(k + 1) - loggamma(n - k + 1) +
+            xlogy(k, p) + xlogy(n - k, 1 - p)
+    else
+        x = -Inf
+    end
+
+    return convert(typeof(float(p)), x)
```
nalimilan: Actually this conversion seems to create problems with dual numbers, as shown by the CI failure. One solution would probably be to do `oftype(float(promote(n, p, k)), x)` (haven't checked).

ericproffitt: this is getting outside my area of expertise, I can make your suggested change if you want, but I can't vouch for its correctness.

nalimilan: At least if tests pass that would be a good sign...

ericproffitt: ok, just made the change

src/basicfuns.jl
oxinabox:
```suggestion
xlogy(x::T, y::T) where {T<:Real} = !iszero(x) ? x * log(y) : zero(log(y))
```

Or even:
```suggestion
xlogy(x, y) = iszero(x) ?  zero(log(y)) : x * log(y)
```

src/basicfuns.jl
```diff
@@ -11,14 +11,14 @@ julia> StatsFuns.xlogx(0)
...
 ```
 """
-xlogx(x::Real) = x > zero(x) ? x * log(x) : zero(log(x))
+xlogx(x::Real) = x != zero(x) ? x * log(x) : zero(log(x))
 
 """
     xlogy(x::Real, y::Real)
 
-Return `x * log(y)` for `y > 0` with correct limit at `x = 0`.
+Return `x * log(y)` for `y ≥ 0` with `x == y = 0` evaluating as limit of xlog(x) as x -> 0 from above.
```
nalimilan: How about being even more explicit?
```suggestion
Return `x * log(y)` for `y ≥ 0`, or `0` when `x == 0` (consistent with the limit
when ```x \\to 0``` from above, including by extension when `y == 0`).
```

ericproffitt: sure, looks good.

test/misc.jl
```diff
@@ -55,3 +55,16 @@ end
...
     end
 end
+
+@testset "binomlogpdf" begin
+    @testset "identity" begin
+        for p in [rand(50); 0; 1]
+            for n in 0:50
+                for k in [-50:50; randn(10)]
+                    # testing genetic Julia binompdf against Rmath version.
+                    @test binomlogpdf(n, Float32(p), k) ≈ log(binompdf(n, p, k))
```
nalimilan: Tests fail as the tolerance is too small by default. I suspect this will actually work:
```suggestion
                    @test Float32(binomlogpdf(n, Float32(p), k)) == Float32(log(binompdf(n, p, k)))
```
Can you check that it's OK? Otherwise you can just specify a reasonable tolerance manually.

ericproffitt: Unfortunately this actually still fails.

however if I use `≈` instead of `==`, it works I think. Although the test dumps some odd output when `k` is not an integer.

Anyways, I made the appropriate change.

src/distrs/binom.jl
```diff
@@ -16,4 +16,13 @@ import .RFunctions:
...
-binomlogpdf(n::Real, p::Real, k::Real) = -log1p(n) - logbeta(n - k + 1, k + 1) + k * log(p) + (n - k) * log1p(-p)
+function binomlogpdf(n::Real, p::Real, k::Real)
+    if isinteger(k) & (zero(k) <= k <= n)
+        x = loggamma(n + 1) - loggamma(k + 1) - loggamma(n - k + 1) +
+            xlogy(k, p) + xlogy(n - k, 1 - p)
+    else
+        x = -Inf
+    end
+
+    return convert(float(promote_type(typeof(n), typeof(p), typeof(k))), x)
```
andreasnoack: Generally, it's better to just compute as much of the general base as possible, i.e. without risking exceptions, and then use the type of that result as the return type. It would be something like
```julia
_lpdf = xlogy(k, p) + xlogy(n - k, 1 - p)
if isinteger(k) & (zero(k) <= k <= n)
    return _lpdf + loggamma(n + 1) - loggamma(k + 1) - loggamma(n - k + 1)
else
    return oftype(_lpdf, -Inf)
end
```
Another thing, by computing `1-p` explicitly, we are losing some precision when `n` is very large. Maybe we need a `xlog1py` or we should just special case the corner case here in this function and continue to use `log1p`.

test/misc.jl
```diff
@@ -55,3 +55,16 @@ end
...
     end
 end
+
+@testset "binomlogpdf" begin
+    @testset "identity" begin
+        for p in [rand(50); 0; 1]
+            for n in 0:50
+                for k in [-50:50; randn(10)]
+                    # testing genetic Julia binompdf against Rmath version.
+                    @test Float32(binomlogpdf(n, Float32(p), k)) ≈ Float32(log(binompdf(n, p, k)))
```
andreasnoack:
```suggestion
                    @test Float32(binomlogpdf(n, Float32(p), k)) ≈ Float32(logbinompdf(n, p, k))
```
will still use the Rmath version and be less prone to overflow so you can test your version for larger `n` and `k`.

src/distrs/binom.jl
nalimilan: Is that really better? I mean, it's much more complex than `float` and it's still not completely right. If we go to that level of complexity, let's do it 100% correctly by also calling `loggamma(one(n-k))`?

andreasnoack: Yes. It's better to avoid `promote_type` and just use a computed value of the correct type. I can't come up with a case where `xlogy` and `loggamma` would produce different types.

nalimilan: Fine with me -- but I find it really annoying that we have to rely on soft rules like that... :-/

andreasnoack: Yes. I've also been annoyed by this. I've been wondering if you could have some kind of language support for branches that enforces the same return type.

nalimilan: Yes, something like `@sametype if` would be nice. But that would only work if inference is able to figure out the type, and AFAIK we shouldn't rely on it. What we could rely on is the fact that for `Number` variables, `one` is defined: the macro could compute the first branch with all variables set to `one(var)`, and then convert the result of the second branch to the type of the result. We probably need a more specific name for that macro...

nalimilan:
@andreasnoack Are you OK with the PR?

andreasnoack:
I'd still prefer that we rely on the promotion in `log` instead of using `promote_type` and `float` explicitly. the former is more general.

nalimilan:
@ericproffitt Do you want to apply what @andreasnoack proposed? BTW, the `xlogx` changes have just been done by another PR so that can be left out of this one.

ericproffitt:
Apologies, I've been unable to set aside enough time to revisit this PR.

Is there someone else who can make the appropriate changes? Otherwise I will revisit when I'm able to.

andreasnoack:
It looks like the xlog(x/y) part has been fixed in versions in https://github.com/JuliaStats/LogExpFunctions.jl and `binomlogpdf` looks like it was fixed by https://github.com/JuliaStats/StatsFuns.jl/pull/113 so I'm closing.

PR #93, opened 2020-02-27
Fix `gamma` argument names

### LINKED ISSUE #92: Argument names of `gammalogpdf` are inconsistent among methods
```julia
help?> Distributions.gammalogpdf
  No documentation found.

  StatsFuns.RFunctions.gammalogpdf is a Function.

  # 2 methods for generic function "gammalogpdf":
  [1] gammalogpdf(α::Union{Float64, Int64}, β::Union{Float64, Int64}, x::Union{Float64, Int64}) in StatsFuns.RFunctions at /home/vargonis/.julia/packages/StatsFuns/dT2Je/src/rmath.jl:73
  [2] gammalogpdf(k::Real, θ::Real, x::Number) in StatsFuns at /home/vargonis/.julia/packages/StatsFuns/dT2Je/src/distrs/gamma.jl:19
```
The two methods do seem to be computing the same thing, though, so it is not a bug. It is the argument names of the first one that violate conventions, and that can induce to errors if one relies on the misleading information they convey.

nalimilan:
`α` and `β` do not match the parameterization that we actually, and these names were inconsistent with the generic method.

Fixes #92.

### DIFF
--- a/src/rmath.jl
+++ b/src/rmath.jl
@@ -116,7 +116,7 @@ end
 @import_rmath binom binom n p
 @import_rmath chisq chisq k
 @import_rmath f fdist ν1 ν2
-@import_rmath gamma gamma α β
+@import_rmath gamma gamma k θ
 @import_rmath hyper hyper ms mf n
 @import_rmath norm norm μ σ
 @import_rmath nbinom nbinom r p


nalimilan:
Note that `beta` uses α and β, but AFAICT that's the only parameterization.

PR #95, opened 2020-05-19
add logsubexp(x,y)

cossio:
This is useful in some contexts. Other libraries have this, for example, https://www.tensorflow.org/probability/api_docs/python/tfp/math/log_sub_exp

If this gets merged please tag a release (I bumped the version here).


src/basicfuns.jl
```diff
@@ -11,15 +11,20 @@ julia> StatsFuns.xlogx(0)
...
 """
     xlogy(x::Real, y::Real)
 
 Return `x * log(y)` for `y > 0` with correct limit at `x = 0`.
 """
-xlogy(x::T, y::T) where {T<:Real} = x > zero(T) ? x * log(y) : zero(log(x))
-xlogy(x::Real, y::Real) = xlogy(promote(x, y)...)
+function xlogy(x, y)
+    result = x * log(y)
+    ifelse(x > zero(x), result, zero(result))
```
simonbyrne, line 26: it does seem a little bit strange to compute the `log` if you're not going to use it.

cossio: The only case where I don't use it is when `x = 0`, which I think is rare. The upside is that this is branch-free.

test/basicfuns.jl
```diff
@@ -88,15 +88,26 @@ end
...
             @test logsumexp(arguments) ≡ result
         end
     end
+    let nan_cases = [[NaN, 9.0], [NaN, Inf], [NaN, -Inf]] # NaN propagation
+        for args in nan_cases
+            @test isnan(logaddexp(args...))
+            @test isnan(logsubexp(args...))
+            @test isnan(logsumexp(args))
+        end
+    end
```
simonbyrne, line 103: This isn't that many cases, so it might be easier just to write them all out.

src/basicfuns.jl
```diff
@@ -196,16 +201,15 @@ end
...
 """
-function logaddexp(x::T, y::T) where T<:Real
-    # x or y is  NaN  =>  NaN
-    # x or y is +Inf  => +Inf
-    # x or y is -Inf  => other value
-    isfinite(x) && isfinite(y) || return max(x,y)
-    x > y ? x + log1p(exp(y - x)) : y + log1p(exp(x - y))
-end
-logaddexp(x::Real, y::Real) = logaddexp(promote(x, y)...)
+logaddexp(x::Real, y::Real) = max(x,y) + log1pexp(-ifelse(x == y, zero(x - y), abs(x - y)))
```
simonbyrne, line 204: this is a lot for a one line function. Also would be good to add a comment as for the reason for the `ifelse`.

test/basicfuns.jl
cossio: Ok, done. I was just trying to be consistent with the pattern of the tests above, but I agree with you.

src/basicfuns.jl
simonbyrne: Wait, is this function even valid for `x < 0`? If not, we should mention that in the docstring.

src/basicfuns.jl
cossio: The `ifelse` is for the case `x = y = Inf`, for which we want `x - y` to be 0. I rewrote this and should be clearer now.

src/basicfuns.jl
cossio: Also this way we get an exception if `y < 0` (even if `x = 0`), which was not the case before. That's a feature I think.

cossio: Oh yes! Had a lapsus. This should error for `x < 0` and/or `y < 0`. I fixed it.

cossio: `xlogx(x)` is only defined for `x >= 0`. But `xlogy(x,y)` I think could take negative `x`? There is no reason to impose the artificial constraint `x >= 0`, right ?

cossio: I just checked that scipy allows negative `x`, https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.xlogy.html

cossio: I added a test for negative `x` to make sure we keep allowing it in the future.

cossio:
Merge?

.gitignore
```diff
@@ -1,2 +1,4 @@
 *.jl.cov
 *.jl.mem
+
```
nalimilan, line 3:
```suggestion
```

src/basicfuns.jl
```diff
@@ -11,15 +11,20 @@ julia> StatsFuns.xlogx(0)
 0.0
 ```
 """
-xlogx(x::Real) = x > zero(x) ? x * log(x) : zero(log(x))
+function xlogx(x)
```
nalimilan, line 14: Are there any other functions where we allow any type? At least I'd use `::Number` here. Also the docstring above must be synced. Same below.

src/basicfuns.jl
```diff
@@ -11,15 +11,20 @@ julia> StatsFuns.xlogx(0)
...
+    result = x * log(x)
+    ifelse(iszero(x), zero(result), result)
```
nalimilan, line 16: Can you update the docstring to reflect more precisely what the function does now? For complex numbers negative values are supported now. Though I'm not sure we should really allow them. Same below.

src/basicfuns.jl
```diff
@@ -196,16 +201,20 @@ end
...
 
-Base.@deprecate logsumexp(x::Real, y::Real) logaddexp(x,y)
+Base.@deprecate logsumexp(x::Real, y::Real) logaddexp(x, y)
+
+"""
+    logsubexp(x, y)
+
+Return `log(abs(e^x - e^y))`, preserving numerical accuracy.
+"""
+logsubexp(x::Real, y::Real) = max(x, y) + log1mexp(-abs(x - y))
```
nalimilan, line 217: Is there no need to handle Inf as in `logaddexp`?

src/basicfuns.jl
```diff
@@ -196,16 +201,20 @@ end
...
 Return `log(exp(x) + exp(y))`, avoiding intermediate overflow/undeflow, and handling non-finite values.
 """
-function logaddexp(x::T, y::T) where T<:Real
-    # x or y is  NaN  =>  NaN
-    # x or y is +Inf  => +Inf
-    # x or y is -Inf  => other value
-    isfinite(x) && isfinite(y) || return max(x,y)
-    x > y ? x + log1p(exp(y - x)) : y + log1p(exp(x - y))
+function logaddexp(x::Real, y::Real)
+    # ensure Δ = 0 if x = y = Inf
```
nalimilan, line 205: Did you change the behavior here?

cossio: No. The behavior is identical. But the new code has no branches.

src/basicfuns.jl
cossio: No. Here it is okay that `logsubexp(Inf,Inf)` returns `NaN`. I test for that explicitly.

src/basicfuns.jl
cossio: If there is no real reason to disallow non-zero complex values, I don't see why we want to place the artificial restriction. scipy's xlogy allows complex values with negative real parts.

src/basicfuns.jl
cossio: Ok, `::Number` seems fine.

src/basicfuns.jl
cossio: Done. Note that for correctness I only have to check zero / non-zero `x` since that's the only condition here. But I added the tests so as to make explicit that we support this and so that it doesn't change inadvertently in the future.

test/basicfuns.jl
```diff
@@ -9,14 +9,19 @@ using StatsFuns, Test
     @test iszero(xlogy(0, 1))
     @test xlogy(2, 3) ≈ 2.0 * log(3.0)
     @test_throws DomainError xlogy(1, -1)
-    # we allow negative `x`, https://github.com/JuliaStats/StatsFuns.jl/pull/95#discussion_r427558736
-    @test xlogy(-2, 3) == -xlogy(2, 3)
     @test isnan(xlogy(NaN, 2))
     @test isnan(xlogy(2, NaN))
+
+    # Since we allow complex/negative values, test for them. See comments in:
```
nalimilan, line 15: Sorry to be a pain, but I'm also concerned with corner cases like those involving a complex and a zero, as well as `Complex(Inf, NaN)` and so on.

cossio: For all those cases it should just return whatever `x * log(y)` turns out to give (which is the correct answer). The only special case is `x = 0` and y != NaN`. Anyway I added all the tests tests I could think of (if you have any special case in mind I can add it as well).

cossio:
@nalimilan Thank you for taking a look!
Nothing left to merge? Also please tag a release with this.

nalimilan:
After considering this again, I [found out](https://www.quora.com/What-is-the-limit-of-z-log-z-as-z-to-0-for-the-complex-multi-valued-natural-logarithm-function-Can-I-choose-a-path-for-z-to-0-that-spirals-slowly-enough-to-prevent-the-limit-from-existing-or-is-it-just-0) that `x log x` has no limit in 0 for complex numbers. So the docstring isn't correct for them. How about just mentioning that we return zero for zero (for both `xlogx` and `xlogy`, like Numpy does? We could still mention that it's consistent with the limit for reals. See also the previous discussion we had about this (I just found it again!): https://github.com/JuliaStats/StatsFuns.jl/pull/87#discussion_r362096613

cossio:
Good point @nalimilan. Done.

src/basicfuns.jl
```diff
@@ -4,22 +4,32 @@
 """
     xlogx(x::Real)
```
nalimilan, line 5: The signature still needs to be updated here.

Also below `x = 0` isn't correct for complex numbers, better say "if x is zero".

cossio: Right, thanks again for noticing! See https://github.com/JuliaStats/StatsFuns.jl/pull/96

PR #96, opened 2020-05-22
fix docstrings of xlogx and xlogy

cossio:
cc @nalimilan

### DIFF
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -17,7 +17,7 @@ function xlogx(x::Number)
 end
 
 """
-    xlogy(x, y)
+    xlogy(x::Number, y::Number)
 
 Compute `x * log(y)`, returning zero if `x` is zero.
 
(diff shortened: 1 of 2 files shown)


src/basicfuns.jl
```diff
@@ -2,9 +2,9 @@
 
 # scalar functions
 """
-    xlogx(x::Real)
+    xlogx(x)
```
nalimilan, line 5:
```suggestion
    xlogx(x::Number)
```

src/basicfuns.jl
```diff
@@ -17,9 +17,9 @@ function xlogx(x::Number)
 end
 
 """
-    xlogy(x::Real, y::Real)
+    xlogy(x, y)
```
nalimilan, line 20:
```suggestion
    xlogy(x::Number, y::Number)
```

cossio: Thanks!

src/basicfuns.jl
cossio: Thanks!

cossio:
`README.md` was out of date so I updated it here.

nalimilan:
Thanks!

Issue #91, opened 2020-02-18
logsumexp for streaming data


cossio:
Consider adding a version of `logsumexp` for streaming data. I found this here:

http://www.nowozin.net/sebastian/blog/streaming-log-sum-exp-computation.html

which includes the following Julia code:

```
function logsumexp_stream(X)
    alpha = -Inf
    r = 0.0
    for x = X
        if x <= alpha
            r += exp(x - alpha)
        else
            r *= exp(alpha - x)
            r += 1.0
            alpha = x
        end
    end
    log(r) + alpha
end
```

It has the advantage that a single pass over the data is required.

Pinging @tpapp in case this should go into https://github.com/tpapp/LogExpFunctions.jl.

I can make a PR myself if people agree into adding this.

nalimilan:
Interesting. Have you benchmarked the two approaches?

cossio:
@nalimilan The original blog post has some benchmarks:

```
n = 10_000_000
X = 500.0*randn(n)

logsumexp_naive(X), logsumexp_batch(X), logsumexp_stream(X)
# (Inf,2686.7659554831052,2686.7659554831052)

julia> @btime logsumexp_batch(X)
92.695 ms (3 allocations: 76.29 MiB)
2600.631695542992

julia> @btime log(sum(exp.(X)))
136.538 ms (6 allocations: 76.29 MiB)
Inf

julia> @btime logsumexp_stream(X)
29.611 ms (1 allocation: 16 bytes)
2600.631695542992
```

Issue #75, opened 2019-10-10
logsumexp for Tuples


cossio:
`logsumexp(X::Tuple)` currently falls back to `reduce(logaddexp, X)`, which is slower and numerically less stable than the array fallback.

Since we can compute the `maximum` of a `Tuple`, can we use the array fallback here too?

Related: https://github.com/JuliaStats/StatsFuns.jl/issues/63

nalimilan:
Yes, probably.

Issue #63, opened 2018-12-19
logsumexp fallback for Base.Iterators.Filter


konkam:
Except for AbstractArrays, `logsumexp` falls back to `reduce(logaddexp, X)`, see: https://github.com/JuliaStats/StatsFuns.jl/blob/316cebb69d4e28e6a4b17d72eee112e61c3b9b77/src/basicfuns.jl#L191

Reduction to adding each element sequentially is numerically less stable than the `sum` function, the issue is mentioned here for instance: https://discourse.julialang.org/t/julia-equivalent-of-pythons-fsum-for-floating-point-summation/17785/27

I have encountered this fallback when using logsumexp on a iterator with a filtering condition:

```
julia> logsumexp(i for i in range(-500, stop = 10, length = 1000) if true) - logsumexp(range(-500, stop = 10, length = 1000))
-1.7763568394002505e-15
```

This may not be a problem related only to logsumexp, maybe the iterator with a filtering condition should also be dealt with as an AbstractArray

nalimilan:
Unfortunately I'm not sure we can do much about this at the moment. First, Julia itself doesn't use the more accurate pairwise summation algorithm for non-arrays (https://github.com/JuliaLang/julia/issues/30421), so we can't compute the precise sum. Then, we need to compute the maximum before computing the sum, which requires doing two passes over the data. But for some iterables, computing each element may be costly or even not possible.

I guess we could have a special case for generators at least, assuming that doing two passes won't be too slow. Or an argument to choose between the fast less accurate method and the possibly slower but accurate one.

konkam:
Ok, many thanks for the reply and the pointer to the similar issue.

devmotion:
This was fixed by https://github.com/JuliaStats/StatsFuns.jl/pull/97, with StatsFuns 0.9.6 I get
```julia
julia> logsumexp(i for i in range(-500, stop = 10, length = 1000) if true) - logsumexp(range(-500, stop = 10, length = 1000))
0.0
```

nalimilan:
Cool!

PR #97, opened 2020-06-08
Add onepass algorithm for logsumexp

devmotion:
This PR fixes https://github.com/JuliaStats/StatsFuns.jl/issues/91.

I used the streaming algorithm for the computation of logsumexp described in http://www.nowozin.net/sebastian/blog/streaming-log-sum-exp-computation.html in one of my projects lately, and I noticed that there is already an open issue for adding it to StatsFuns.

I reran the benchmark mentioned in the issue and got:
```julia
julia> using StatsFuns, BenchmarkTools, Random

julia> Random.seed!(100);

julia> n = 10_000_000;

julia> X = 500 .* randn(n);

julia> @btime logsumexp($X)
  94.999 ms (0 allocations: 0 bytes)
2697.6523838563817

julia> @btime StatsFuns.logsumexp_onepass($X)
  32.815 ms (0 allocations: 0 bytes)
2697.6523838563817

julia> f(X) = reduce(logaddexp, X)
f (generic function with 1 method)

julia> @btime f($X)
  162.519 ms (0 allocations: 0 bytes)
2697.6523838563817
```

Currently `logsumexp_onepass` is used as the fallback for `logsumexp` but not exported. I'm not sure if it should be exported as well. Additionally, maybe it would be beneficial to use the one-pass algorithms even for arrays above a certain size, since it seems to perform better in the benchmark above. Some additional benchmarks suggest that this might be the case even for smaller arrays:
```julia
julia> n = 10;

julia> Random.seed!(100);

julia> X = 500 .* randn(n);

julia> @btime logsumexp($X)
  84.612 ns (0 allocations: 0 bytes)
1016.2819596629249

julia> @btime StatsFuns.logsumexp_onepass($X)
  87.535 ns (0 allocations: 0 bytes)
1016.2819596629249

julia> @btime f($X)
  184.664 ns (0 allocations: 0 bytes)
1016.2819596629249

julia> n = 100;

julia> Random.seed!(100);

julia> X = 500 .* randn(n);

julia> @btime logsumexp($X)
  633.000 ns (0 allocations: 0 bytes)
1786.3178411042175

julia> @btime StatsFuns.logsumexp_onepass($X)
  654.025 ns (0 allocations: 0 bytes)
1786.3178411042175

julia> @btime f($X)
  1.337 μs (0 allocations: 0 bytes)
1786.3178411042175

julia> n = 1_000;

julia> Random.seed!(100);

julia> X = 500 .* randn(n);

julia> @btime logsumexp($X)
  6.698 μs (0 allocations: 0 bytes)
1786.3178411042175

julia> @btime StatsFuns.logsumexp_onepass($X)
  3.854 μs (0 allocations: 0 bytes)
1786.3178411042175

julia> @btime f($X)
  11.460 μs (0 allocations: 0 bytes)
1786.3178411042175
```

__Edit:__ This PR also fixes the example in https://github.com/JuliaStats/StatsFuns.jl/issues/63:

```julia
julia> logsumexp(i for i in range(-500, stop = 10, length = 1000) if true) - logsumexp(range(-500, stop = 10, length = 1000))
0.0
```

Additionally, it addresses (and fixes?) https://github.com/JuliaStats/StatsFuns.jl/issues/75 (but maybe we would still want the two-pass algorithm instead?).

### DIFF
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -261,22 +261,25 @@ end
 function _logsumexp_onepass(X)
     # fallback for empty collections
     isempty(X) && return log(sum(X))
-    xmax, r = _logsumexp_onepass(X, Base.IteratorEltype(X))
-    return xmax + log1p(r)
+    return _logsumexp_onepass_result(_logsumexp_onepass_reduce(X, Base.IteratorEltype(X)))
 end
 
-# initial element is required by CUDA (otherwise we could remove this method)
-function _logsumexp_onepass(X, ::Base.HasEltype)
+# function barrier for reductions with single element and without initial element
+_logsumexp_onepass_result(x) = float(x)
+_logsumexp_onepass_result((xmax, r)::Tuple) = xmax + log1p(r)
+
+# iterables with known element type
+function _logsumexp_onepass_reduce(X, ::Base.HasEltype)
     # do not perform type computations if element type is abstract
     T = eltype(X)
-    isconcretetype(T) || return _logsumexp_onepass(X, Base.EltypeUnknown())
+    isconcretetype(T) || return _logsumexp_onepass_reduce(X, Base.EltypeUnknown())
 
     FT = float(T)
     return reduce(_logsumexp_onepass_op, X; init=(FT(-Inf), zero(FT)))
 end
 
-# without initial element (without CUDA support we could always use this method)
-_logsumexp_onepass(X, ::Base.EltypeUnknown)::Tuple = reduce(_logsumexp_onepass_op, X)
+# iterables without known element type
+_logsumexp_onepass_reduce(X, ::Base.EltypeUnknown) = reduce(_logsumexp_onepass_op, X)
 
 ## Reductions for one-pass algorithm: avoid expensive multiplications if numbers are reduced
 
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -227,11 +227,27 @@ logsubexp(x::Real, y::Real) = max(x, y) + log1mexp(-abs(x - y))
 Compute `log(sum(exp, X))` in a numerically stable way that avoids intermediate over- and
 underflow.
 
-`X` should be an iterator of real numbers.
+`X` should be an iterator of real numbers. The result is computed using a single pass over
+the data.
 
-See also: [`logsumexp_onepass`](@ref)
+# References
+
+[Sebastian Nowozin: Streaming Log-sum-exp Computation.](http://www.nowozin.net/sebastian/blog/streaming-log-sum-exp-computation.html)
 """
 logsumexp(X) = logsumexp_onepass(X)
+
+"""
+    logsumexp(X::AbstractArray{<:Real}[; dims=:])
+
+Compute `log.(sum(exp.(X); dims=dims))` in a numerically stable way that avoids
+intermediate over- and underflow.
+
+If `dims = :`, then the result is computed using a single pass over the data.
+
+# References
+
+[Sebastian Nowozin: Streaming Log-sum-exp Computation.](http://www.nowozin.net/sebastian/blog/streaming-log-sum-exp-computation.html)
+"""
 logsumexp(X::AbstractArray{<:Real}; dims=:) = _logsumexp(X, dims)
 
 _logsumexp(X::AbstractArray{<:Real}, ::Colon) = logsumexp_onepass(X)
@@ -241,19 +257,6 @@ function _logsumexp(X::AbstractArray{<:Real}, dims)
     return u .+ log.(sum(exp.(X .- u); dims=dims))
 end
 
-"""
-    logsumexp_onepass(X)
-
-Compute `log(sum(exp, X))` in a numerically stable way that avoids intermediate under- and
-overflow.
-
-In contrast to [`logsumexp`](@ref) the result is computed using a single pass over the data.
-`X` should be an iterator of real numbers.
-
-# References
-
-[Sebastian Nowozin: Streaming Log-sum-exp Computation.](http://www.nowozin.net/sebastian/blog/streaming-log-sum-exp-computation.html)
-"""
 function logsumexp_onepass(X)
     # fallback for empty collections
     isempty(X) && return log(sum(X))
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -207,7 +207,7 @@ end
 Return `log(exp(x) + exp(y))`, avoiding intermediate overflow/undeflow, and handling non-finite values.
 """
 function logaddexp(x::Real, y::Real)
-    # ensure Δ = 0 if x = y = Inf
+    # ensure Δ = 0 if x = y = ± Inf
     Δ = ifelse(x == y, zero(x - y), abs(x - y))
     max(x, y) + log1pexp(-Δ)
 end
@@ -224,14 +224,14 @@ logsubexp(x::Real, y::Real) = max(x, y) + log1mexp(-abs(x - y))
 """
     logsumexp(X)
 
-Compute `log(sum(exp, X))`, evaluated avoiding intermediate overflow/undeflow.
+Compute `log(sum(exp, X))` in a numerically stable way that avoids intermediate over- and
+underflow.
 
 `X` should be an iterator of real numbers.
+
+See also: [`logsumexp_onepass`](@ref)
 """
-function logsumexp(X)
-    isempty(X) && return log(sum(X))
-    reduce(logaddexp, X)
-end
+logsumexp(X) = logsumexp_onepass(X)
 function logsumexp(X::AbstractArray{T}; dims=:) where {T<:Real}
     # Do not use log(zero(T)) directly to avoid issues with ForwardDiff (#82)
     u = reduce(max, X, dims=dims, init=oftype(log(zero(T)), -Inf))
@@ -246,6 +246,45 @@ function logsumexp(X::AbstractArray{T}; dims=:) where {T<:Real}
     end
 end
 
+"""
+    logsumexp_onepass(X)
+
+Compute `log(sum(exp, X))` in a numerically stable way that avoids intermediate under- and
+overflow.
+
+In contrast to [`logsumexp`](@ref) the result is computed using a single pass over the data.
+`X` should be an iterator of real numbers.
+
+# References
+
+[Sebastian Nowozin: Streaming Log-sum-exp Computation.](http://www.nowozin.net/sebastian/blog/streaming-log-sum-exp-computation.html)
+"""
+function logsumexp_onepass(X)
+    isempty(X) && return log(sum(X))
+
+    # initialize maximum value and accumulated sum for the first iterate
+    x, state = iterate(X)
+    xmax = x
+    r = exp(zero(x))
+    r_one = r
+
+    # for all other iterates
+    while (next = iterate(X, state)) !== nothing
+        x, state = next
+
+        # update maximum value and accumulated sum
+        if x < xmax
+            r += exp(x - xmax)
+        elseif x > xmax
+            r = r_one + r * exp(xmax - x)
+            xmax = x
+        else # ensure finite values if x = xmax = ± Inf
+            r += r_one
+        end
+    end
+
+    return xmax + log(r)
+end
 
 """
     softmax!(r::AbstractArray, x::AbstractArray)
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -237,7 +237,7 @@ the data.
 logsumexp(X) = logsumexp_onepass(X)
 
 """
-    logsumexp(X::AbstractArray{<:Real}[; dims=:])
+    logsumexp(X::AbstractArray{<:Real}; dims=:)
 
 Compute `log.(sum(exp.(X); dims=dims))` in a numerically stable way that avoids
 intermediate over- and underflow.
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -268,6 +268,10 @@ end
 
 # with initial element: required by CUDA
 function _logsumexp_onepass(X, ::Base.HasEltype)
+    # do not perform type computations if element type is abstract
+    T = eltype(X)
+    isconcretetype(T) || return _logsumexp_onepass(X, Base.EltypeUnknown())
+
     # compute initial element
     FT = float(eltype(X))
     init = (FT(-Inf), zero(FT))
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -238,15 +238,7 @@ _logsumexp(X::AbstractArray{<:Real}, ::Colon) = logsumexp_onepass(X)
 function _logsumexp(X::AbstractArray{<:Real}, dims)
     # Do not use log(zero(eltype(X))) directly to avoid issues with ForwardDiff (#82)
     u = reduce(max, X, dims=dims, init=oftype(log(zero(eltype(X))), -Inf))
-    u isa AbstractArray || isfinite(u) || return float(u)
-    let u=u # avoid https://github.com/JuliaLang/julia/issues/15276
-        # TODO: remove the branch when JuliaLang/julia#31020 is merged.
-        if u isa AbstractArray
-            u .+ log.(sum(exp.(X .- u); dims=dims))
-        else
-            u + log(sum(x -> exp(x-u), X))
-        end
-    end
+    return u .+ log.(sum(exp.(X .- u); dims=dims))
 end
 
 """
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -266,7 +266,7 @@ function logsumexp_onepass(X)
     return xmax + log(r)
 end
 
-# with initial element: required by CUDA
+# initial element is required by CUDA (ideally we would never use this method)
 function _logsumexp_onepass(X, ::Base.HasEltype)
     # do not perform type computations if element type is abstract
     T = eltype(X)
@@ -277,13 +277,12 @@ function _logsumexp_onepass(X, ::Base.HasEltype)
     init = (FT(-Inf), zero(FT))
     r_one = one(FT)
 
-    # perform single pass over the data
     return mapreduce(_logsumexp_onepass_op, X; init=init) do x
         return float(x), r_one
     end
 end
 
-# without initial element
+# without initial element (ideally we would always use this method)
 function _logsumexp_onepass(X, ::Base.EltypeUnknown)
     return mapreduce(_logsumexp_onepass_op, X) do x
         _x = float(x)
@@ -291,17 +290,19 @@ function _logsumexp_onepass(X, ::Base.EltypeUnknown)
     end
 end
 
+# all inputs are provided as floating point numbers
+# `r1` and `r2` are one if `xmax1` and `xmax2` are new elements (no partial sums)
 function _logsumexp_onepass_op((xmax1, r1)::T, (xmax2, r2)::T) where {T<:Tuple}
     if xmax1 < xmax2
         xmax = xmax2
         a = exp(xmax1 - xmax2)
-        r = r2 + ifelse(isone(r1), a, r1 * a)
+        r = r2 + (isone(r1) ? a : r1 * a) # avoid expensive multiplication for new elements
     elseif xmax1 > xmax2
         xmax = xmax1
         a = exp(xmax2 - xmax1)
-        r = r1 + ifelse(isone(r2), a, r2 * a)
+        r = r1 + (isone(r2) ? a : r2 * a) # avoid expensive multiplication for new elements
     else # ensure finite values if x = xmax = ± Inf
-        xmax = ifelse(isnan(xmax1), xmax1, xmax2)
+        xmax = isnan(xmax1) ? xmax1 : xmax2
         r = r1 + r2
     end
 
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -232,9 +232,12 @@ underflow.
 See also: [`logsumexp_onepass`](@ref)
 """
 logsumexp(X) = logsumexp_onepass(X)
-function logsumexp(X::AbstractArray{T}; dims=:) where {T<:Real}
-    # Do not use log(zero(T)) directly to avoid issues with ForwardDiff (#82)
-    u = reduce(max, X, dims=dims, init=oftype(log(zero(T)), -Inf))
+logsumexp(X::AbstractArray{<:Real}; dims=:) = _logsumexp(X, dims)
+
+_logsumexp(X::AbstractArray{<:Real}, ::Colon) = logsumexp_onepass(X)
+function _logsumexp(X::AbstractArray{<:Real}, dims)
+    # Do not use log(zero(eltype(X))) directly to avoid issues with ForwardDiff (#82)
+    u = reduce(max, X, dims=dims, init=oftype(log(zero(eltype(X))), -Inf))
     u isa AbstractArray || isfinite(u) || return float(u)
     let u=u # avoid https://github.com/JuliaLang/julia/issues/15276
         # TODO: remove the branch when JuliaLang/julia#31020 is merged.
@@ -260,30 +263,51 @@ In contrast to [`logsumexp`](@ref) the result is computed using a single pass ov
 [Sebastian Nowozin: Streaming Log-sum-exp Computation.](http://www.nowozin.net/sebastian/blog/streaming-log-sum-exp-computation.html)
 """
 function logsumexp_onepass(X)
+    # fallback for empty collections
     isempty(X) && return log(sum(X))
 
-    # initialize maximum value and accumulated sum for the first iterate
-    x, state = iterate(X)
-    xmax = x
-    r = exp(zero(x))
-    r_one = r
-
-    # for all other iterates
-    while (next = iterate(X, state)) !== nothing
-        x, state = next
-
-        # update maximum value and accumulated sum
-        if x < xmax
-            r += exp(x - xmax)
-        elseif x > xmax
-            r = r_one + r * exp(xmax - x)
-            xmax = x
-        else # ensure finite values if x = xmax = ± Inf
-            r += r_one
-        end
-    end
+    # perform single pass over the data
+    xmax, r = _logsumexp_onepass(X, Base.IteratorEltype(X))
 
     return xmax + log(r)
+end
+
+# with initial element: required by CUDA
+function _logsumexp_onepass(X, ::Base.HasEltype)
+    # compute initial element
+    FT = float(eltype(X))
+    init = (FT(-Inf), zero(FT))
+    r_one = one(FT)
+
+    # perform single pass over the data
+    return mapreduce(_logsumexp_onepass_op, X; init=init) do x
+        return float(x), r_one
+    end
+end
+
+# without initial element
+function _logsumexp_onepass(X, ::Base.EltypeUnknown)
+    return mapreduce(_logsumexp_onepass_op, X) do x
+        _x = float(x)
+        return _x, one(_x)
+    end
+end
+
+function _logsumexp_onepass_op((xmax1, r1)::T, (xmax2, r2)::T) where {T<:Tuple}
+    if xmax1 < xmax2
+        xmax = xmax2
+        a = exp(xmax1 - xmax2)
+        r = r2 + ifelse(isone(r1), a, r1 * a)
+    elseif xmax1 > xmax2
+        xmax = xmax1
+        a = exp(xmax2 - xmax1)
+        r = r1 + ifelse(isone(r2), a, r2 * a)
+    else # ensure finite values if x = xmax = ± Inf
+        xmax = ifelse(isnan(xmax1), xmax1, xmax2)
+        r = r1 + r2
+    end
+
+    return xmax, r
 end
 
 """
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -258,7 +258,6 @@ function logsumexp_onepass(X)
     # fallback for empty collections
     isempty(X) && return log(sum(X))
 
-    # perform single pass over the data
     xmax, r = _logsumexp_onepass(X, Base.IteratorEltype(X))
 
     return xmax + log(r)
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -234,7 +234,7 @@ the data.
 
 [Sebastian Nowozin: Streaming Log-sum-exp Computation.](http://www.nowozin.net/sebastian/blog/streaming-log-sum-exp-computation.html)
 """
-logsumexp(X) = logsumexp_onepass(X)
+logsumexp(X) = _logsumexp_onepass(X)
 
 """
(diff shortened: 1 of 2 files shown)


ararslan:
Seems like we could just use this algorithm for `logsumexp` instead of what's currently there without needing a separate function for it. The benchmarks you posted seem to suggest that would be beneficial; any reason not to do that?

devmotion:
The algorithm doesn't handle reductions along certain dimensions, so I guess it would only replace the case with `dims = :` in a straightforward way. Additionally, it seems that for `n = 100` elements the two-pass is still slightly faster (it seems the one-pass algorithm takes over at some point between 100 and 1000 elements in my benchmarks).

devmotion:
Bump :)

devmotion:
I addressed this issue in a separate PR.

dmbates:
> I addressed this issue in a separate PR.

Thank you.

nalimilan:
AFAICT your algorithm is only very slightly slower for small arrays, and much faster for large arrays, right? So better use it instead of the existing ` reduce(logaddexp, X)`. (We don't usually provide two functions for different algorithms doing the same thing.)

src/basicfuns.jl
```diff
@@ -224,14 +224,14 @@ logsubexp(x::Real, y::Real) = max(x, y) + log1mexp(-abs(x - y))
...
 `X` should be an iterator of real numbers.
+
+See also: [`logsumexp_onepass`](@ref)
 """
-function logsumexp(X)
-    isempty(X) && return log(sum(X))
-    reduce(logaddexp, X)
-end
+logsumexp(X) = logsumexp_onepass(X)
 function logsumexp(X::AbstractArray{T}; dims=:) where {T<:Real}
```
nalimilan, line 235: When `dims=:`,  `logsumexp_onepass` should be called. The standard way to achieve this is to have `_logsumexp(X, ::Colon)` and `_logsumexp(X::AbstractArray, dims)`, and call them from `logsumexp(X; dims=:) = _logsumexp(X, dims)`.

Also there's no need to define `logsumexp_onepass`, just make it a `_logsumexp` method.

src/basicfuns.jl
```diff
@@ -246,6 +246,45 @@ function logsumexp(X::AbstractArray{T}; dims=:) where {T<:Real}
...
+    isempty(X) && return log(sum(X))
+
+    # initialize maximum value and accumulated sum for the first iterate
+    x, state = iterate(X)
+    xmax = x
+    r = exp(zero(x))
+    r_one = r
+
+    # for all other iterates
+    while (next = iterate(X, state)) !== nothing
```
nalimilan, line 272: FWIW, if you're looking for the last bit of performance, it could be faster to have another method specific to `AbstractArray` using `@inbounds @simd for x in X`. Often it makes a difference.

nalimilan:
@simonbyrne wrote the previous implementation at https://github.com/JuliaStats/StatsFuns.jl/pull/45, so I'd like him to confirm it's OK, in particular for custom arrays like `CuArrays`, where `reduce` might have been faster.

I also wonder why `reduce(logaddexp, X)` is so slow.

devmotion:
> I also wonder why `reduce(logaddexp, X)` is so slow.

`reduce(logaddexp, X)` calls `log1pexp` in every step, which in the example above should call `log1p(exp(x))` in roughly half of the steps. Surely this will be computationally more expensive than just calling `exp(...)` in every step (and one additional `log(...)` call in the end).

devmotion:
> @simonbyrne wrote the previous implementation at #45, so I'd like him to confirm it's OK, in particular for custom arrays like CuArrays, where reduce might have been faster.

Calling `logsumexp` on a `CuArray` will not fall back to the `reduce` version but rather on the version that is implemented for `AbstractArray`. In https://github.com/JuliaStats/StatsFuns.jl/pull/45, the `reduce` version was just added as a generic fallback but it seems the main change to enable the use of `CuArray`s was to remove the indexing from the implementation for `AbstractArray`. Hence I'm not convinced that `logsumexp_onepass` should be used as a replacement for `logsumexp(x::AbstractArray; dims = :)`.

The one-pass algorithm could be used as a replacement for the generic `reduce` fallback without defining an explicit `logsumexp_onepass` function, but then it would not be possible to use it with `AbstractArray`s.

Therefore I created a separate `logsumexp_onepass` function that is used as the generic fallback of `logsumexp`.

nalimilan:
Unfortunately, ArrayInterface currently depends on Requires, which increases the load time significantly and is therefore probably not OK for this package. We really need this kind of traits in a lightweight package containing only basic definitions. I guess in the meantime is would be OK to use `logsumexp_onepass` for `Array` only, and leave a TODO.

Though another more interesting solution would be to adapt the one-pass algorithm to use `mapreduce` instead. That way, the optimized `CuArray` methods (which distribute work in parallel over multiple blocks) would be used, and we would get the best of both worlds. AFAICT this should be possible, as the operation applied to each element is associative, right?

devmotion:
> Though another more interesting solution would be to adapt the one-pass algorithm to use mapreduce instead. That way, the optimized CuArray methods (which distribute work in parallel over multiple blocks) would be used, and we would get the best of both worlds. AFAICT this should be possible, as the operation applied to each element is associative, right?

Hmm I don't see how this would be possible with `mapreduce`: the transformation applied to each element depends on the result of the reduction so far, so it is not clear to me how one could define a function of the current element only for the `map` part of `mapreduce`. AFAICT it is possible to rewrite the implementation using `foldl` or `foldr`. However, since every reduction step results in a tuple of the current maximum and the sum of the exponentials of all elements so far, divided by the exponential of the largest element so far, the reduction is not associative (for `foldl` the reduction has the signature `f(::Tuple{<:Real,<:Real}, ::Real)` whereas for `foldr` it is `f(::Real, ::Tuple{<:Real,<:Real})`), and hence one can't use `reduce`.

nalimilan:
Yeah, the algorithm currently can only handle one new element at a time. But I wonder whether it would be possible to use a trick in order to keep track of whether the reduction operation is currently working on a single element or on a partial sum of multiple elements. Something like
```julia
mapreduce(x -> (x, 1), ((x, i), (y, j)) -> (i == 1 && j == 1 ? : ... : (i == 1 ? ... : (j == 1 ? ... : ...)), i+j), v)
```
AFAICT that would work if we have a formula to combine two partial sums into a single one. Is that the case? :-)

If that doesn't work, let's go with the special-casing of `Array`.

devmotion:
Yes, I guess that could work :+1: The reduction of two partial sums seems to be very similar to the currently implemented version AFAICT. I'll see if it works and benchmark both variants.

devmotion:
Hmm it works mathematically but it seems `mapreduce` is quite slow:
```julia
julia> using StatsFuns, BenchmarkTools, Random

julia> Random.seed!(100);

julia> n = 10_000_000;

julia> X = 500 .* randn(n);

julia> @btime logsumexp($X)
  106.407 ms (0 allocations: 0 bytes)
2570.6633157197443

julia> @btime StatsFuns.logsumexp_onepass($X)
  40.213 ms (0 allocations: 0 bytes)
2570.6633157197443

julia> @btime StatsFuns.logsumexp_onepass_reduce($X)
  80.846 ms (0 allocations: 0 bytes)
2570.6633157197443

julia> f(X) = reduce(logaddexp, X)
f (generic function with 1 method)

julia> @btime f($X)
  167.125 ms (0 allocations: 0 bytes)
2570.6633157197443
```

nalimilan:
Ah, too bad. Though it's a bit faster than the current implementation. I wonder what would be the result for `CuArray`. `mapreduce` also has the potential of using multiple threads once that support is added.

Have you tried forcing `@inbounds` just in case?

devmotion:
> Have you tried forcing @inbounds just in case?

There's no need for `@inbounds` in the `mapreduce` version it seems. I updated the comment above with the implementation that I tested.

nalimilan:
Sorry I meant `@inline`.

devmotion:
> mapreduce also has the potential of using multiple threads once that support is added.

Yeah I agree, that would be great. Also it seems by reducing the partial sums it would be trivial to implement `logsumexp` with (parallel) transducers.

devmotion:
Unfortunately, the timings are the same even if I add an `@inline` to the reduction function.

nalimilan:
OK, too bad. Maybe it would be faster if you passed `i` and `j` as in my example so that you can identify the most common case where one of the argument is a single value. Then in that case you would be able to do `if x < xmax; r += exp(x - xmax)` as in `logsumexp_onepass`. Indeed `mapreduce` IIRC is going to call `op(r, f(x[i]))` most of the time (as opposed to combining two partial sums, which will only happen once for each block), so that's where performance matters and getting rid of one multiply operation should make a difference.

devmotion:
But then we're basically back at foldl and foldr since the reduction would not be associative anymore, it seems? Since intuitively the most obvious thing would be to not map the entries at all if we want to know that we process a new element.

nalimilan:
No, I'm just suggesting to use a slightly simpler expression when `i == 1` or `j == 1`, in the hope that branch prediction will make that fast. But you would still have the other branches as in your current code.

devmotion:
I saw your example above but I still don't see how this could work with integer-valued `i` and `j`. During the reduction we need both the current maximum `xmax` and the remainder part `r` of the partial sum - but `r` is not integer-valued, so depending on if it is a single element or not, one would get different input types. But in this case we could just define the reduction for combinations of real numbers and tuples instead of artificially constructed tuples, it seems? Hence the more natural approach would be to just use `foldl` or `foldr` directly (or possibly `reduce` if we define the reduction for both combinations of real numbers and tuples).

devmotion:
For instance, this code for `foldl`
```julia
logsumexp_onepass_foldl_reduction(::Nothing, x) = x, float(one(x))
function logsumexp_onepass_foldl_reduction((xmax, r)::Tuple, x)
    if x < xmax
        return xmax, r + exp(x - xmax)
    elseif x > xmax
        return x, float(one(x)) + r * exp(xmax - x)
    else
        return xmax, float(one(x)) + r
    end
end

function logsumexp_onepass_foldl(X)

    xmax, r = foldl(logsumexp_onepass_foldl_reduction, X; init=nothing)

    return xmax + log(r)
end
```
is as efficient as the current implementation in this PR, it seems:
```julia
julia> @btime StatsFuns.logsumexp_onepass($X)
  40.060 ms (0 allocations: 0 bytes)
2570.6633157197443

julia> @btime StatsFuns.logsumexp_onepass_foldl($X)
  39.900 ms (0 allocations: 0 bytes)
2570.6633157197443
```

devmotion:
BTW I just tried to run the `logsumexp_onepass_reduce` version on a GPU using `X = 500 .* cu(randn(n))`. It fails with the error message
```julia
julia> @btime StatsFuns.logsumexp_onepass_reduce($X)
ERROR: GPUArrays.jl needs to know the neutral element for your operator `logsumexp_onepass_reduction`.
Please pass it as an explicit argument to (if possible), or register it
globally your operator by defining `GPUArrays.neutral_element(::typeof(logsumexp_onepass_reduction), T)`.
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] neutral_element(::Function, ::Type{T} where T) at /home/davwi492/.julia/packages/GPUArrays/eVYIC/src/host/mapreduce.jl:13
...
```

nalimilan:
> I saw your example above but I still don't see how this could work with integer-valued `i` and `j`. During the reduction we need both the current maximum `xmax` and the remainder part `r` of the partial sum - but `r` is not integer-valued, so depending on if it is a single element or not, one would get different input types. But in this case we could just define the reduction for combinations of real numbers and tuples instead of artificially constructed tuples, it seems? Hence the more natural approach would be to just use `foldl` or `foldr` directly (or possibly `reduce` if we define the reduction for both combinations of real numbers and tuples).

I was thinking of passing three arguments: the current maximum, the partial sum, and the number of elements that have been summed so far (or just a Boolean indicating whether the partial sum is actually a single element). But you're right that relying on dispatch might also work -- not sure which approach would be the fastest.

`foldl` doesn't have much advantage compared with a loop since it's not parallelizable.

The error with `CuArray` might be fixable by passing `init` to `mapreduce`. But I must say I have zero knowledge of CUDA. :-)

If that's too complex, let's go with a special-case for `Array` and use `mapreduce` only for other `AbstractArray`s. What do you think?

devmotion:
The following seems to work both for regular `Array`s and `CuArray`s, and is (almost) as fast as the current implementation in this PR:
```julia
function _logsumexp_onepass_op((xmax1, r1)::T, (xmax2, r2)::T) where {T<:Tuple}
    if xmax1 < xmax2
        xmax = xmax2
        a = exp(xmax1 - xmax2)
        r = r2 + ifelse(isone(r1), a, r1 * a)
    elseif xmax1 > xmax2
        xmax = xmax1
        a = exp(xmax2 - xmax1)
        r = r1 + ifelse(isone(r2), a, r2 * a)
    else # ensure finite values if x = xmax = ± Inf
        xmax = xmax1
        r = r1 + r2
    end

    return xmax, r
end

function logsumexp_onepass_reduce(X)

    T = float(eltype(X))
    init = (T(-Inf), zero(T))
    r_one = one(T)
    xmax, r = mapreduce(_logsumexp_onepass_op, X; init=init) do x
        float(x), r_one
    end

    return xmax + log(r)
end
```

CPU:
```julia
julia> using StatsFuns, BenchmarkTools, Random

julia> n = 10_000_000;

julia> X = 500 .* randn(n);

julia> @btime logsumexp($X)
  105.061 ms (0 allocations: 0 bytes)
2617.9174023177598

julia> @btime StatsFuns.logsumexp_onepass($X)
  42.841 ms (0 allocations: 0 bytes)
2617.9174023177598

julia> @btime StatsFuns.logsumexp_onepass_reduce($X)
  42.866 ms (0 allocations: 0 bytes)
2617.9174023177598
```

GPU:
```julia
julia> using StatsFuns, BenchmarkTools, CUDA, Random

julia> n = 10_000_000;

julia> X = 500 .* cu(randn(n));

julia> @btime logsumexp($X)
  271.088 μs (146 allocations: 4.09 KiB)
2696.5413f0

julia> @btime StatsFuns.logsumexp_onepass_reduce($X)
  212.357 μs (71 allocations: 2.02 KiB)
2696.5413f0

julia> n = 10;

julia> X = 500 .* cu(randn(n));

julia> @btime logsumexp($X)
  27.410 μs (78 allocations: 2.19 KiB)
685.3175f0

julia> @btime StatsFuns.logsumexp_onepass_reduce($X)
  13.590 μs (37 allocations: 1.06 KiB)
685.3175f0

julia> n = 100;

julia> X = 500 .* cu(randn(n));

julia> @btime logsumexp($X)
  27.572 μs (78 allocations: 2.19 KiB)
1339.7037f0

julia> @btime StatsFuns.logsumexp_onepass_reduce($X)
  16.828 μs (37 allocations: 1.06 KiB)
1339.7037f0

julia> n = 1_000;

julia> X = 500 .* cu(randn(n));

julia> @btime logsumexp($X)
  45.219 μs (146 allocations: 4.09 KiB)
1668.9694f0

julia> @btime StatsFuns.logsumexp_onepass_reduce($X)
  20.218 μs (37 allocations: 1.06 KiB)
1668.9694f0
```

So I assume we might want to use the `logsumexp_onepass_reduce` implementation as a generic fallback for `logsumexp(X)` and for `_logsumexp(X::AbstractArray{<:Real}, ::Colon)` where `logsumexp(X::AbstractArray{<:Real}; dims=:) = _logsumexp(X, dims)`?

nalimilan:
Fantastic. Let’s do that.

src/basicfuns.jl
```diff
@@ -224,17 +224,20 @@ logsubexp(x::Real, y::Real) = max(x, y) + log1mexp(-abs(x - y))
 """
     logsumexp(X)
 
-Compute `log(sum(exp, X))`, evaluated avoiding intermediate overflow/undeflow.
+Compute `log(sum(exp, X))` in a numerically stable way that avoids intermediate over- and
+underflow.
 
 `X` should be an iterator of real numbers.
+
+See also: [`logsumexp_onepass`](@ref)
```
nalimilan, line 232: Move all documentation relevant for end users to this docstring, and remove the one for 
`logsumexp_onepass`, which is internal and not supposed to be used directly.
```suggestion
```

src/basicfuns.jl
```diff
@@ -246,6 +249,66 @@ function logsumexp(X::AbstractArray{T}; dims=:) where {T<:Real}
...
+
+# References
+
+[Sebastian Nowozin: Streaming Log-sum-exp Computation.](http://www.nowozin.net/sebastian/blog/streaming-log-sum-exp-computation.html)
+"""
+function logsumexp_onepass(X)
+    # fallback for empty collections
+    isempty(X) && return log(sum(X))
+
+    # perform single pass over the data
```
nalimilan, line 269: I think as this point it's clear enough that this function does a single pass. :-)
```suggestion
```

src/basicfuns.jl
```diff
@@ -246,6 +249,66 @@ function logsumexp(X::AbstractArray{T}; dims=:) where {T<:Real}
...
+        _x = float(x)
+        return _x, one(_x)
+    end
+end
+
+function _logsumexp_onepass_op((xmax1, r1)::T, (xmax2, r2)::T) where {T<:Tuple}
+    if xmax1 < xmax2
+        xmax = xmax2
+        a = exp(xmax1 - xmax2)
+        r = r2 + ifelse(isone(r1), a, r1 * a)
```
nalimilan, line 300: Maybe add a comment explaining that the `isone` check is there for performance? BTW, I'm surprised it makes things faster when using `ifelse`, since both branches are supposed to be evaluated in that case (I'd have used `?`).

test/basicfuns.jl
```diff
@@ -97,8 +98,10 @@ end
     @test logaddexp(10002, 10003) ≈ 10000 + logaddexp(2.0, 3.0)
 
     @test logsumexp([1.0, 2.0, 3.0])          ≈ 3.40760596444438
+    @test logsumexp_onepass([1.0, 2.0, 3.0])  ≈ 3.40760596444438
```
nalimilan, line 101: No need to test this now that it's the default. Same below: better test the public-facing function.

src/basicfuns.jl
```diff
@@ -224,17 +224,20 @@ logsubexp(x::Real, y::Real) = max(x, y) + log1mexp(-abs(x - y))
...
-    # Do not use log(zero(T)) directly to avoid issues with ForwardDiff (#82)
-    u = reduce(max, X, dims=dims, init=oftype(log(zero(T)), -Inf))
+logsumexp(X) = logsumexp_onepass(X)
+logsumexp(X::AbstractArray{<:Real}; dims=:) = _logsumexp(X, dims)
+
+_logsumexp(X::AbstractArray{<:Real}, ::Colon) = logsumexp_onepass(X)
+function _logsumexp(X::AbstractArray{<:Real}, dims)
+    # Do not use log(zero(eltype(X))) directly to avoid issues with ForwardDiff (#82)
+    u = reduce(max, X, dims=dims, init=oftype(log(zero(eltype(X))), -Inf))
     u isa AbstractArray || isfinite(u) || return float(u)
```
nalimilan, line 241: Am I right that `u` will always be an `AbstractArray` now?

src/basicfuns.jl
```diff
@@ -246,6 +249,66 @@ function logsumexp(X::AbstractArray{T}; dims=:) where {T<:Real}
...
+    # fallback for empty collections
+    isempty(X) && return log(sum(X))
+
+    # perform single pass over the data
+    xmax, r = _logsumexp_onepass(X, Base.IteratorEltype(X))
+
+    return xmax + log(r)
+end
+
+# with initial element: required by CUDA
```
nalimilan, line 275: I guess it's also required by `Array` when the input is empty, right?

src/basicfuns.jl
devmotion: I guess so. Just checked that even if all dimensions are specified such as in `sum(rand(3, 4); dims=(1,2))` a matrix is returned.

src/basicfuns.jl
devmotion: No, empty inputs are handled by `log(sum(X))`, as before. Actually we never want to use this method apart for GPU arrays - the type calculations are doomed to fail e.g. for not concretely typed arrays and hence it is much safer to not rely on them and instead just compute the output. I'm not sure why the type calculations in https://github.com/JuliaGPU/GPUArrays.jl/blob/356c5fe3f83f76f8139b4edf5536cbd08d48da7f/src/host/mapreduce.jl#L49-L52 fail, I can check if this can be fixed from our side. Alternatively, maybe another package such as NNlib should provide a GPU-compatible overload of StatsFuns.logsumexp?

nalimilan: OK. They fail because `neutral_element` is only defined for a few particular functions. Another approach would be to use the type of the first element, since we already handle the case where the array is empty. That way we wouldn't need two different methods depending on whether the eltype is known.

Given that it seems doable, it would be nice to have something that works for all arrays without another package having to overload the method.

src/basicfuns.jl
nalimilan: OK so we can just remove the `u isa AbstractArray` checks.

src/basicfuns.jl
devmotion: Runs on the GPU fail even before this point due to the type calculations. IMO it is just bad that GPUArrays requires us to provide an initial or neutral element. We really never want to do that, regardless if there is a first element or not. Retrieving the first element is bad for any stateful iterator, and is a scalar indexing operation on the GPU (which leads to warnings when calling the function - and we can not remove them without depending on CUDA). Additionally, the docstring of `mapreduce` says that
```julia
It is unspecified whether init is used for non-empty collections.
```
so there is really no point in providing `init` here - apart from GPU compatibility. The problem is that we can not restrict our implementation with the `init` hack to only GPU arrays without taking a dependency on CUDA. Hence I don't think it is completely unreasonable to provide a GPU-compatible implementation somewhere else. An alternative would be to not use the one-pass algorithm as the default algorithm (for arrays).

src/basicfuns.jl
```diff
@@ -227,46 +227,40 @@ logsubexp(x::Real, y::Real) = max(x, y) + log1mexp(-abs(x - y))
...
-            u .+ log.(sum(exp.(X .- u); dims=dims))
-        else
-            u + log(sum(x -> exp(x-u), X))
-        end
-    end
-end
 
 """
-    logsumexp_onepass(X)
+    logsumexp(X::AbstractArray{<:Real}[; dims=:])
```
nalimilan, line 240: No need for brackets, they are implicit when there's a default value:
```suggestion
    logsumexp(X::AbstractArray{<:Real}; dims=:)
```

test/basicfuns.jl
```diff
@@ -1,4 +1,5 @@
 using StatsFuns, Test
+using StatsFuns: logsumexp_onepass
```
nalimilan, line 2:
```suggestion
```

test/basicfuns.jl
```diff
@@ -137,6 +138,11 @@ end
     @test isnan(logsumexp([NaN, 9.0]))
     @test isnan(logsumexp([NaN, Inf]))
     @test isnan(logsumexp([NaN, -Inf]))
+
+    # issue #63
+    a = logsumexp(i for i in range(-500, stop = 10, length = 1000) if true)
+    b = logsumexp(range(-500, stop = 10, length = 1000))
+    @test a == b
```
nalimilan, line 145: Even without the `if` you get a `Generator`, which is enough to test the method:
```suggestion
    # logsumexp with general iterables (issue #63)
    x = logsumexp(range(-500, stop = 10, length = 1000))
    @test logsumexp(i for i in x) == logsumexp(x)
```

src/basicfuns.jl
nalimilan: Retrieving the first element should be OK for any `AbstractArray` (contrary to some iterables). It's too bad that CUDA prints warnings in that case.

Let's keep what you have now: since the method only accepts arrays with `T<:Real` element type, `float(T)` will (almost) certainly work.

src/basicfuns.jl
nalimilan: Have you noticed this one?

nalimilan (commented):
Just a few remaining details.

test/basicfuns.jl
devmotion:
```suggestion
    # logsumexp with general iterables (issue #63)
    xs = range(-500, stop = 10, length = 1000)
    @test logsumexp(x for x in xs) == logsumexp(xs)
```

src/basicfuns.jl
devmotion: Yes, I just wanted to postpone any changes to the algorithm until the general layout (regarding CUDA etc) was clear.

src/basicfuns.jl
```diff
@@ -224,28 +224,90 @@ logsubexp(x::Real, y::Real) = max(x, y) + log1mexp(-abs(x - y))
...
 """
-function logsumexp(X)
+logsumexp(X::AbstractArray{<:Real}; dims=:) = _logsumexp(X, dims)
+
+_logsumexp(X::AbstractArray{<:Real}, ::Colon) = logsumexp_onepass(X)
+function _logsumexp(X::AbstractArray{<:Real}, dims)
+    # Do not use log(zero(eltype(X))) directly to avoid issues with ForwardDiff (#82)
+    u = reduce(max, X, dims=dims, init=oftype(log(zero(eltype(X))), -Inf))
+    return u .+ log.(sum(exp.(X .- u); dims=dims))
+end
```
nalimilan, line 258: Actually, why don't we use `mapreduce(_logsumexp_onepass_op, ..., dims=dims)` here too? That should also be faster, and avoid the intermediate allocation of `u` and `exp.(X .- u)`.

src/basicfuns.jl
```diff
@@ -224,28 +224,96 @@ logsubexp(x::Real, y::Real) = max(x, y) + log1mexp(-abs(x - y))
...
+[Sebastian Nowozin: Streaming Log-sum-exp Computation.](http://www.nowozin.net/sebastian/blog/streaming-log-sum-exp-computation.html)
+"""
+logsumexp(X::AbstractArray{<:Real}; dims=:) = _logsumexp(X, dims)
+
+_logsumexp(X::AbstractArray{<:Real}, ::Colon) = _logsumexp_onepass(X)
+function _logsumexp(X::AbstractArray{<:Real}, dims)
+    # Do not use log(zero(eltype(X))) directly to avoid issues with ForwardDiff (#82)
+    FT = float(eltype(X))
+    xmax_r = reduce(_logsumexp_onepass_op, X; dims=dims, init=(FT(-Inf), zero(FT)))
+    return @. first(xmax_r) + log1p(last(xmax_r))
```
nalimilan, line 258: Slightly simpler:
```suggestion
    xmax, r = reduce(_logsumexp_onepass_op, X; dims=dims, init=(FT(-Inf), zero(FT)))
    return xmax .+ log1p.(r)
```

src/basicfuns.jl
```diff
@@ -224,28 +224,96 @@ logsubexp(x::Real, y::Real) = max(x, y) + log1mexp(-abs(x - y))
...
+    # do not perform type computations if element type is abstract
+    T = eltype(X)
+    isconcretetype(T) || return _logsumexp_onepass(X, Base.EltypeUnknown())
+
+    FT = float(T)
+    return reduce(_logsumexp_onepass_op, X; init=(FT(-Inf), zero(FT)))
+end
+
+# without initial element (without CUDA support we could always use this method)
+_logsumexp_onepass(X, ::Base.EltypeUnknown)::Tuple = reduce(_logsumexp_onepass_op, X)
```
nalimilan, line 279: Is `Tuple` really useful here (it's not there for the other method)?
```suggestion
_logsumexp_onepass(X, ::Base.EltypeUnknown) = reduce(_logsumexp_onepass_op, X)
```

nalimilan (approved):
Cool. Looks good if you confirm the latest version is as fast as the previous one. I have two stylistic comments.

src/basicfuns.jl
devmotion: No, that's incorrect. `xmax_r` is a `AbstractArray{<:Tuple}`.

src/basicfuns.jl
devmotion: Yes, it is needed to make the function inferrable (the compiler infers the return type as `Union{Float64,Tuple{Float64,Float64}}`). I added it since `X` is always non-empty but I just noticed that this is incorrect for iterables with a single element - this needs an additional branch or an initial value it seems which is quite annoying...

test/basicfuns.jl
```diff
@@ -137,6 +137,10 @@ end
     @test isnan(logsumexp([NaN, 9.0]))
     @test isnan(logsumexp([NaN, Inf]))
     @test isnan(logsumexp([NaN, -Inf]))
+
+    # logsumexp with general iterables (issue #63)
+    xs = range(-500, stop = 10, length = 1000)
+    @test logsumexp(x for x in xs) == logsumexp(xs)
 end
```
nalimilan, line 144: @stevengj mentioned [here](https://discourse.julialang.org/t/fast-logsumexp/22827/7) that underflow can also be a problem for `logsumexp` implementations, e.g. with `[1e-20, log(1e-20)]`. It seems that your PR handles this correctly (returning `1.99...e-20`, while the previous one incorrectly returned `1.0e-20`. Maybe add a test above around line 105 to cover this case?

src/basicfuns.jl
nalimilan: Does it make a difference for performance? Otherwise, maybe that's not a real issue.

test/basicfuns.jl
devmotion: Sure, I added one.

src/basicfuns.jl
devmotion: I added a function barrier instead.

devmotion:
Same benchmark as above gives now
```julia
julia> @btime StatsFuns.logsumexp($X)
  37.203 ms (0 allocations: 0 bytes)
2570.6633157197443
```
on my computer.

devmotion:
For the calls with `dims` I ran
```julia
julia> using StatsFuns, BenchmarkTools, Random

julia> Random.seed!(100);

julia> n = 100_000;

julia> X = 500 .* randn(n, 100);
```
and got on this PR
```julia
julia> @btime StatsFuns.logsumexp($X; dims=2);
  99.331 ms (8 allocations: 2.29 MiB)

julia> @btime StatsFuns.logsumexp($X; dims=1);
  63.163 ms (2 allocations: 2.64 KiB)

julia> @btime StatsFuns.logsumexp($X; dims=(1,2));
  66.100 ms (6 allocations: 352 bytes)
```
whereas master yields
```julia
julia> @btime StatsFuns.logsumexp($X; dims=2);
  117.644 ms (16 allocations: 78.58 MiB)

julia> @btime StatsFuns.logsumexp($X; dims=1);
  86.431 ms (5 allocations: 76.30 MiB)

julia> @btime StatsFuns.logsumexp($X; dims=(1,2));
  72.502 ms (13 allocations: 76.29 MiB)
```

nalimilan:
Thanks @devmotion!

PR #108, opened 2021-02-17
Reexport LogExpFunctions

### LINKED ISSUE #46: RFC: move log/exp variants to a separate package
I am thinking of the functions in [src/basicfuns.jl](https://github.com/JuliaStats/StatsFuns.jl/blob/e66dd973650c375bc1739c820e5b96bb5bd000a8/src/basicfuns.jl).

In contrast to distribution functions, these are all implemented in pure Julia, require no external dependencies, and thus have no license issues. In fact, while they are used in statistics, they are not even "statistical" functions since other fields make ample use of them. Cf #20; these functions are useful for the end user directly, without requiring an interface like Distributions.jl.

I propose `LogExpFunctions.jl` as a package name. If the discussion suggests this makes sense, I would be happy to do a PR.

devmotion:
Fixes #46.

The basic `log`/`exp` functions were all moved to LogExpFunctions (see https://github.com/JuliaStats/StatsFuns.jl/issues/46#issuecomment-769868836). This PR removes the duplicate implementations in StatsFuns and reexports the definitions in LogExpFunctions.


tpapp:
Incidentally, I initiated the repo transfer to JuliaStats, but got no reply. If it somehow expired, I am happy to do it again.

src/StatsFuns.jl
```diff
@@ -233,10 +208,37 @@ export
     logmvbeta,          # logarithm of multivariate beta function
     lstirling_asym
 
+# reexports
+using LogExpFunctions
+using LogExpFunctions: loghalf, logtwo, logπ, log2π, log4π, softplus, invsoftplus
```
devmotion, line 213: @tpapp I guess it would be reasonable to export them in LogExpFunctions as well?

cossio:
What's holding this up?

nalimilan (approved):
Looks good, but I haven't seen any notification regarding the transfert request. If you make me an owner of the repo I'll do it.

Also, could you preserve the git history from StatsFuns in the new repo? That would be useful.

tpapp:
@nalimilan: apparently I need permissions on JuliaStats to make the transfer, can you add me?

I don't know how to preserve the git history.

nalimilan:
> @nalimilan: apparently I need permissions on JuliaStats to make the transfer, can you add me?

Yes the person doing the transfer needs to be an owner of both source and destination. I've sent you an invitation.

> I don't know how to preserve the git history.

See explanations here for example: https://github.com/JuliaLang/julia/pull/31395#issuecomment-515488104

tpapp:
Thanks. I made the transfer, but currently have no time for the git history operation. If anyone wants to do it, please do not hesitate.

devmotion:
I opened a PR to the registry as well (https://github.com/JuliaRegistries/General/pull/30590), following the instructions in https://github.com/JuliaRegistries/General#how-do-i-transfer-a-package-to-an-organization-or-another-user.

nalimilan:
OK, I've run `git-filter-repo --path src/basicfuns.jl --path test/basicfuns.jl` on the StatsFuns repo, fetched it as a branch in LogSumExpFunctions, rebased master on top of it, and pushed it to [the StatsFuns-rebased branch](https://github.com/JuliaStats/LogExpFunctions.jl/tree/StatsFuns-rebased).

But that's not ideal as another commit synced against StatsFuns later so the diff is backwards. So I pushed another [StatsFuns-rebased2 branch](https://github.com/JuliaStats/LogExpFunctions.jl/compare/StatsFuns-rebased2) in which I squashed the few initial commits up to the sync. This gives [a much cleaner diff](https://github.com/JuliaStats/LogExpFunctions.jl/commit/19a2c9c6e078a4422ba7e1c7b9b29f6f35e79fdf#diff-3effc2d68a83d383489434663c6d00967cf5a4e0394d32ca421b5541b9f949dd). What do you think? I guess we should keep existing commits somewhere since releases point to them, but we could start from there for the new main?

devmotion:
I am not sure what's the best approach to keep/transfer some git history. Since there are already releases of LogExpFunctions in the general registry that point to commits in the repo it seems we shouldn't touch the existing commits? So even though I would prefer if the history could be kept, I am not sure if it can be done in a clean way. Maybe one could add a commit that removes all contents in the LogExpFunctions repo, then add a copy of the StatsFuns files, and afterwards re-apply the current commits in the LogExpFunctions repo rebased on the copy (but maybe without the sync commit)?

devmotion:
I went for a simpler but less sophisticated approach (i.e. more work :smile:): I extracted the git history of the relevant files with `git filter-repo` and merged it with the updated docstrings in LogExpFunctions. I opened a PR: https://github.com/JuliaStats/LogExpFunctions.jl/pull/11

nalimilan (approved):

PR #110, opened 2021-03-28
Add Rmath 0.7

ViralBShah:
Perhaps ok to remove the older Rmath versions from the compat list?


ViralBShah:
Hehe - that is what CompatHelper is for.

PR #111, opened 2021-03-28
Use github ci

ViralBShah:


nalimilan:
For the record, this PR mistakenly dropped CI on Julia 1.0 without bumping the required version in Project.toml, and later PRs broke pre-1.3 Julia compatibility without anybody noticing. We should be very careful about this. See https://github.com/JuliaStats/StatsFuns.jl/pull/138.

PR #112, opened 2021-04-06
Julia implementations for `poiscdf` and `poislogcdf`

torfjelde:
This PR adds Julia implementations for `poiscdf` and `poislogcdf`.

I'm a bit uncertain where tests for these should go, so I would appreciate some advice on that:)

### DIFF
--- a/src/distrs/pois.jl
+++ b/src/distrs/pois.jl
@@ -23,7 +23,8 @@ function poislogcdf(λ::Number, x::Number)
     return loggamma(floor(x + 1), float(λ)) - logfactorial(floor(x))
 end
 
-poiscdf(λ::Number, x::Number) = exp(poislogcdf(λ, x))
+poiscdf(λ::Number, x::Number; precision=0) = last(gamma_inc(floor(x + 1), λ, precision))
+poisccdf(λ::Number, x::Number; precision=0) = first(gamma_inc(floor(x + 1), λ, precision))
 
 #=
 function poislogpdf(λ::Union{Float32,Float64}, x::Union{Float64,Float32,Integer})
--- a/src/distrs/pois.jl
+++ b/src/distrs/pois.jl
@@ -19,10 +19,8 @@ poislogpdf(λ::T, x::T) where {T <: Real} = xlogy(x, λ) - λ - loggamma(x + 1)
 
 poislogpdf(λ::Number, x::Number) = poislogpdf(promote(float(λ), x)...)
 
-poislogpdf(λ::Number, x::Number) = poislogpdf(promote(float(λ), x)...)
-
 function poislogcdf(λ::Number, x::Number)
-    return SpecialFunctions.loggamma(Int(x) + 1, float(λ)) - SpecialFunctions.logfactorial(Int(x))
+    return loggamma(floor(x + 1), float(λ)) - logfactorial(floor(x))
 end
 
 poiscdf(λ::Number, x::Number) = exp(poislogcdf(λ, x))
--- a/src/distrs/pois.jl
+++ b/src/distrs/pois.jl
@@ -19,6 +19,14 @@ poislogpdf(λ::T, x::T) where {T <: Real} = xlogy(x, λ) - λ - loggamma(x + 1)
 
 poislogpdf(λ::Number, x::Number) = poislogpdf(promote(float(λ), x)...)
 
+poislogpdf(λ::Number, x::Number) = poislogpdf(promote(float(λ), x)...)
+
+function poislogcdf(λ::Number, x::Number)
+    return SpecialFunctions.loggamma(Int(x) + 1, λ) - SpecialFunctions.logfactorial(Int(x))
+end
+
+poiscdf(λ::Number, x::Number) = exp(poislogcdf(λ, x))
+
 #=
 function poislogpdf(λ::Union{Float32,Float64}, x::Union{Float64,Float32,Integer})
     if iszero(λ)
--- a/src/distrs/pois.jl
+++ b/src/distrs/pois.jl
@@ -22,7 +22,7 @@ poislogpdf(λ::Number, x::Number) = poislogpdf(promote(float(λ), x)...)
 poislogpdf(λ::Number, x::Number) = poislogpdf(promote(float(λ), x)...)
 
 function poislogcdf(λ::Number, x::Number)
-    return SpecialFunctions.loggamma(Int(x) + 1, λ) - SpecialFunctions.logfactorial(Int(x))
+    return SpecialFunctions.loggamma(Int(x) + 1, float(λ)) - SpecialFunctions.logfactorial(Int(x))
 end
 
 poiscdf(λ::Number, x::Number) = exp(poislogcdf(λ, x))
--- a/src/distrs/pois.jl
+++ b/src/distrs/pois.jl
@@ -19,12 +19,15 @@ poislogpdf(λ::T, x::T) where {T <: Real} = xlogy(x, λ) - λ - loggamma(x + 1)
 
 poislogpdf(λ::Number, x::Number) = poislogpdf(promote(float(λ), x)...)
 
-function poislogcdf(λ::Number, x::Number)
+function poislogcdf(λ::T, x::T) where {T<:Real}
     return loggamma(floor(x + 1), float(λ)) - logfactorial(floor(x))
 end
+poislogcdf(λ::Number, x::Number) = poislogcdf(promote(float(λ), x)...)
 
-poiscdf(λ::Number, x::Number; precision=0) = last(gamma_inc(floor(x + 1), λ, precision))
-poisccdf(λ::Number, x::Number; precision=0) = first(gamma_inc(floor(x + 1), λ, precision))
+poiscdf(λ::T, x::T; precision=0) where {T<:Real} = last(gamma_inc(floor(x + 1), λ, precision))
+poiscdf(λ::Number, x::Number; precision=0) = poiscdf(promote(float(λ), x)...)
+poisccdf(λ::T, x::T; precision=0) where {T<:Real} = first(gamma_inc(floor(x + 1), λ, precision))
+poisccdf(λ::Number, x::Number; precision=0) = poisccdf(promote(float(λ), x)...)
 
 #=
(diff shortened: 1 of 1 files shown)


src/distrs/pois.jl
```diff
@@ -19,6 +19,14 @@ poislogpdf(λ::T, x::T) where {T <: Real} = xlogy(x, λ) - λ - loggamma(x + 1)
 
 poislogpdf(λ::Number, x::Number) = poislogpdf(promote(float(λ), x)...)
 
+poislogpdf(λ::Number, x::Number) = poislogpdf(promote(float(λ), x)...)
+
+function poislogcdf(λ::Number, x::Number)
+    return SpecialFunctions.loggamma(Int(x) + 1, float(λ)) - SpecialFunctions.logfactorial(Int(x))
+end
+
+poiscdf(λ::Number, x::Number) = exp(poislogcdf(λ, x))
```
devmotion, line 28: To me it seems it would be preferable to define
```suggestion
poiscdf(λ::Number, x::Number) = last(gamma_inc(floor(x + 1), λ))
poisccdf(λ::Number, x::Number) = first(gamma_inc(floor(x + 1), λ))
```

src/distrs/pois.jl
```diff
@@ -19,6 +19,14 @@ poislogpdf(λ::T, x::T) where {T <: Real} = xlogy(x, λ) - λ - loggamma(x + 1)
 
 poislogpdf(λ::Number, x::Number) = poislogpdf(promote(float(λ), x)...)
 
+poislogpdf(λ::Number, x::Number) = poislogpdf(promote(float(λ), x)...)
+
```
devmotion, line 23: It seems this is copied from above?
```suggestion
```

src/distrs/pois.jl
```diff
@@ -19,6 +19,14 @@ poislogpdf(λ::T, x::T) where {T <: Real} = xlogy(x, λ) - λ - loggamma(x + 1)
...
+function poislogcdf(λ::Number, x::Number)
+    return SpecialFunctions.loggamma(Int(x) + 1, float(λ)) - SpecialFunctions.logfactorial(Int(x))
+end
```
devmotion, line 26: It seems a bit restrictive to assume that `x` is an integer:
```suggestion
function poislogcdf(λ::Number, x::Number)
    return loggamma(floor(x + 1), float(λ)) - logfactorial(floor(x))
end
```
Are you sure it is necessary to call `float`?

src/distrs/pois.jl
devmotion: Alternatively, one could also forward the precision setting to SpecialFunctions (https://juliamath.github.io/SpecialFunctions.jl/stable/functions_list/#SpecialFunctions.gamma_inc):
```suggestion
poiscdf(λ::Number, x::Number; precision=0) = last(gamma_inc(floor(x + 1), λ, precision))
poisccdf(λ::Number, x::Number; precision=0) = first(gamma_inc(floor(x + 1), λ, precision))
```

src/distrs/pois.jl
torfjelde: > Are you sure it is necessary to call float?

Nope! I just did because of the code above.

torfjelde: Oh and regarding restricting `Int`: I assumed the methods would be more numerically stable when doing so + couldn't see why you'd call these with non-integer types, i.e. I would think if you write `1.0` you actually mean `1`. 

I'm fine with whatever here though!

src/distrs/pois.jl
torfjelde: Ah yes, this was by mistake! Thanks:)

src/distrs/pois.jl
devmotion: You could want to evaluate the CDF also for e.g. `x=1.5`. This makes sense mathematically and is supported by Rmath.

src/distrs/pois.jl
torfjelde: This doesn't cost you anything, given that it computes both upper and first?

devmotion:
The generic implementations of (log)pdf are compared with Rmath in https://github.com/JuliaStats/StatsFuns.jl/blob/master/test/generic.jl. Similarly, one should check the implementations of (log)cdf.

src/distrs/pois.jl
torfjelde: When you say "mathematically", you mean in the sense that "the mathematical function supports it", not in the sense that P(x \le 1.5) \ne P(x \le 1) for a discrete rv. x, right?

torfjelde: Ah nvm, you added a `floor` function: I am happy with this:)

torfjelde:
> The generic implementations of (log)pdf are compared with Rmath in https://github.com/JuliaStats/StatsFuns.jl/blob/master/test/generic.jl. Similarly, one should check the implementations of (log)cdf.

Cool :+1: But there are some that already have these implemented, e.g. `normcdf`, but there doesn't exist a test similar to `genericcomp` for `cdf` already?

src/distrs/pois.jl
devmotion: Both results sum up to 1, so it should not cost anything. And I assume that otherwise there would be two separate functions in SpecialFunctions.

devmotion:
No, I think it is missing.

src/distrs/pois.jl
torfjelde: > sum up to 1

Doh. Makes sense then:)

src/distrs/pois.jl
```diff
@@ -19,6 +19,13 @@ poislogpdf(λ::T, x::T) where {T <: Real} = xlogy(x, λ) - λ - loggamma(x + 1)
 
 poislogpdf(λ::Number, x::Number) = poislogpdf(promote(float(λ), x)...)
 
+function poislogcdf(λ::Number, x::Number)
+    return loggamma(floor(x + 1), float(λ)) - logfactorial(floor(x))
+end
+
+poiscdf(λ::Number, x::Number; precision=0) = last(gamma_inc(floor(x + 1), λ, precision))
+poisccdf(λ::Number, x::Number; precision=0) = first(gamma_inc(floor(x + 1), λ, precision))
```
torfjelde, line 27: Btw, should `precision` be argument rather than kwarg? @devmotion

devmotion: It doesn't matter but IMO it is a bit more natural as a keyword argument since we don't dispatch on it.

src/distrs/pois.jl
```diff
@@ -19,6 +19,13 @@ poislogpdf(λ::T, x::T) where {T <: Real} = xlogy(x, λ) - λ - loggamma(x + 1)
 
 poislogpdf(λ::Number, x::Number) = poislogpdf(promote(float(λ), x)...)
```
torfjelde, line 20: I think I get why this is done btw: `loggamma(::ForwardDiff.Dual, ::Int)` won't have a diff-rule defined for it.

torfjelde: Actually, it doesn't seem like these have rules defined in DiffRules.jl!

devmotion: There is no DiffRule defined for `loggamma` and `gamma_inc` anyway so it shouldn't matter at the moment :shrug: I added ChainRules definitions to SpecialFunctions (https://github.com/JuliaMath/SpecialFunctions.jl/pull/305) but currently they can't be used with Zygote due to https://github.com/FluxML/Zygote.jl/issues/873.

So can we avoid these conversions if possible for now and only add them later if needed?

torfjelde: I'll have to look at it later, but it seems like there are particular execution branches (depending on the values) which ForwardDiff.jl fails to automatically obtain derivatives for. So when I just tried a couple of values everything seemed to work, but when I let HMC loose on a distribution using this function under the hood, it eventually runs into a `MethodError`.

torfjelde: Haha, man, you're always like one step ahead: I just looked at your PR and that's exactly the "solution" i arrived at literally 10mins ago, i.e. we can only take derivative wrt. second argument but that's also what we care about so w/e.

> There is no DiffRule defined for loggamma and gamma_inc anyway so it shouldn't matter at the moment 

But wouldn't adding those fix this issue? Same as you've done in ChainRules now?

devmotion: Sure, if I used the correct derivatives it should be fine :stuck_out_tongue: AFAICT the convention in DiffRules would be to set the derivatives of the first argument to NaN, if it is not implemented properly.

andreasnoack:
Regarding the testing then see https://github.com/JuliaStats/StatsFuns.jl/pull/113. I think we should just start to use the native implementations unconditionally. Then they'll be tested automatically against the Rmath versions with the current test setup.

src/distrs/pois.jl
andreasnoack: I don't think we should expose the `precision` argument here. The implementation in SpecialFunctions happens to support it but it's not something that is generally available I don't think it will be in demand here. If it is then I think it will be more natural to get the lower precision version as part of a method for `Float32`.

src/distrs/pois.jl
```diff
@@ -19,6 +19,16 @@ poislogpdf(λ::T, x::T) where {T <: Real} = xlogy(x, λ) - λ - loggamma(x + 1)
...
 
+function poislogcdf(λ::T, x::T) where {T<:Real}
+    return loggamma(floor(x + 1), float(λ)) - logfactorial(floor(x))
```
andreasnoack, line 23: Very minor detail, but it seems a little odd to me use `logfactorial` here together with `loggamma`. In particular when we also use `loggamma(x+1)` in the pdf.

andreasnoack:
@torfjelde It would be great to have this one completed. Do you expect to have the cycles to complete it soon?

torfjelde:
> It would be great to have this one completed. Do you expect to have the cycles to complete it soon?

Sorry, been quite busy lately. I'm travelling this weekend so won't be able to do much, but I'll try to get it done early next week if that's okay?

andreasnoack:
No worries. I ended up doing it as part of #113

torfjelde:
Btw, should we just close this issue then?

andreasnoack:
Yeah. Let's do that (and then I'll have to get #113 completed)

PR #113, opened 2021-05-13
Use julia implementations for pdfs and some cdf-like functions

andreasnoack:
I think we should begin to use Julia implementations when they are available. There is is really no need to call out to Rmath when we have implementations available. Eventually, we should then also be able to drop Rmath as a package dependency and just use it for testing although there is still some work to do. However, SpecialFunctions now has some of the more tricky functions available so we can get quite far. This has a conflict with https://github.com/JuliaStats/StatsFuns.jl/pull/112 so it would be good to get than one merged first and I can update this PR afterwards.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -14,6 +14,7 @@ SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 
 [compat]
 ChainRulesCore = "1"
+HypergeometricFunctions = "0.3"
 InverseFunctions = "0.1"
 IrrationalConstants = "0.1"
 LogExpFunctions = "0.3.2"
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ version = "0.9.15"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
+HypergeometricFunctions = "34004b35-14d8-5ef3-9330-4cdb6864b03a"
 InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
 IrrationalConstants = "92d709cd-6900-40b7-9082-c6be49f344b6"
 LogExpFunctions = "2ab3a3ac-af41-5b50-aa03-7779005ae688"
--- a/src/distrs/beta.jl
+++ b/src/distrs/beta.jl
@@ -1,15 +1,17 @@
 # functions related to beta distributions
 
+using HypergeometricFunctions: _₂F₁
+
 # R implementations
 using .RFunctions:
     # betapdf,
     # betalogpdf,
-    betacdf,
-    betaccdf,
-    betalogcdf,
-    betalogccdf,
-    betainvcdf,
-    betainvccdf,
+    # betacdf,
+    # betaccdf,
+    # betalogcdf,
+    # betalogccdf,
+    # betainvcdf,
+    # betainvccdf,
     betainvlogcdf,
     betainvlogccdf
 
@@ -23,3 +25,46 @@ function betalogpdf(α::T, β::T, x::T) where {T<:Real}
     val = xlogy(α - 1, y) + xlog1py(β - 1, -y) - logbeta(α, β)
     return x < 0 || x > 1 ? oftype(val, -Inf) : val
 end
+
+betacdf(α::Float64, β::Float64, x::Float64) = first(beta_inc(α, β, x))
+betacdf(α::Real, β::Real, x::Real) = betacdf(promote(float(α), β, x)...)
+betacdf(k::T, x::T) where T = throw(MethodError(betacdf, (α, β, x)))
+
+betaccdf(α::Float64, β::Float64, x::Float64) = last(beta_inc(α, β, x))
+betaccdf(α::Real, β::Real, x::Real) = betaccdf(promote(float(α), β, x)...)
+betaccdf(k::T, x::T) where T = throw(MethodError(betaccdf, (α, β, x)))
+
+# The log version is currently based on non-log version. When the cdf is very small we shift
+# to an implementation based on the hypergeometric function ₂F₁ to avoid underflow.
+function betalogcdf(α::Float64, β::Float64, x::Float64)
+    p, q = beta_inc(α, β, x)
+    if p < eps(one(p))
+        # see https://dlmf.nist.gov/8.17#E7
+        return -log(α) + α*log(x) + log(_₂F₁(promote(α, 1 - β, α + 1, x)...)) - logbeta(α, β)
+    elseif p <= 0.7
+        return log(p)
+    else
+        return log1p(-q)
+    end
+end
+betalogcdf(α::Real, β::Real, x::Real) = betalogcdf(promote(float(α), β, x)...)
+betalogcdf(k::T, x::T) where T = throw(MethodError(betalogcdf, (α, β, x)))
+
+function betalogccdf(α::Float64, β::Float64, x::Float64)
+    p, q = beta_inc(α, β, x)
+    if q < 0.7
+        return log(q)
+    else
+        return log1p(-p)
+    end
+end
+betalogccdf(α::Real, β::Real, x::Real) = betalogccdf(promote(float(α), β, x)...)
+betalogccdf(α::T, β::T, x::T) where T = throw(MethodError(betalogccdf, (α, β, x)))
+
+betainvcdf(α::Float64, β::Float64, p::Float64) = first(beta_inc_inv(α, β, p, 1 - p))
+betainvcdf(α::Real, β::Real, p::Real) = betainvcdf(promote(float(α), β, p)...)
+betainvcdf(α::T, β::T, p::T) where T = throw(MethodError(betainvcdf, (α, β, p)))
+
+betainvccdf(α::Float64, β::Float64, p::Float64) = first(beta_inc_inv(α, β, 1 - p, p))
+betainvccdf(α::Real, β::Real, p::Real) = betainvccdf(promote(float(α), β, p)...)
+betainvccdf(α::T, β::T, p::T) where T = throw(MethodError(betainvccdf, (α, β, p)))
--- a/src/distrs/beta.jl
+++ b/src/distrs/beta.jl
@@ -32,17 +32,18 @@ betaccdf(α::Real, β::Real, x::Real) = last(beta_inc(α, β, x))
 
 # The log version is currently based on non-log version. When the cdf is very small we shift
 # to an implementation based on the hypergeometric function ₂F₁ to avoid underflow.
-function betalogcdf(α::Real, β::Real, x::Real)
+function betalogcdf(α::T, β::T, x::T) where {T<:Real}
     p, q = beta_inc(α, β, x)
     if p < eps(one(p))
         # see https://dlmf.nist.gov/8.17#E7
-        return -log(α) + α*log(x) + log(_₂F₁(promote(α, 1 - β, α + 1, x)...)) - logbeta(α, β)
+        return -log(α) + xlogy(α, x) + log(_₂F₁(promote(α, 1 - β, α + 1, x)...)) - logbeta(α, β)
     elseif p <= 0.7
         return log(p)
     else
         return log1p(-q)
     end
 end
+betalogcdf(α::Real, β::Real, x::Real) = betalogcdf(promote(α, β, x)...)
 
 function betalogccdf(α::Real, β::Real, x::Real)
     p, q = beta_inc(α, β, x)
--- a/src/distrs/beta.jl
+++ b/src/distrs/beta.jl
@@ -26,17 +26,13 @@ function betalogpdf(α::T, β::T, x::T) where {T<:Real}
     return x < 0 || x > 1 ? oftype(val, -Inf) : val
 end
 
-betacdf(α::Float64, β::Float64, x::Float64) = first(beta_inc(α, β, x))
-betacdf(α::Real, β::Real, x::Real) = betacdf(promote(float(α), β, x)...)
-betacdf(k::T, x::T) where T = throw(MethodError(betacdf, (α, β, x)))
+betacdf(α::Real, β::Real, x::Real) = first(beta_inc(α, β, x))
 
-betaccdf(α::Float64, β::Float64, x::Float64) = last(beta_inc(α, β, x))
-betaccdf(α::Real, β::Real, x::Real) = betaccdf(promote(float(α), β, x)...)
-betaccdf(k::T, x::T) where T = throw(MethodError(betaccdf, (α, β, x)))
+betaccdf(α::Real, β::Real, x::Real) = last(beta_inc(α, β, x))
 
 # The log version is currently based on non-log version. When the cdf is very small we shift
 # to an implementation based on the hypergeometric function ₂F₁ to avoid underflow.
-function betalogcdf(α::Float64, β::Float64, x::Float64)
+function betalogcdf(α::Real, β::Real, x::Real)
     p, q = beta_inc(α, β, x)
     if p < eps(one(p))
         # see https://dlmf.nist.gov/8.17#E7
@@ -47,10 +43,8 @@ function betalogcdf(α::Float64, β::Float64, x::Float64)
         return log1p(-q)
     end
 end
-betalogcdf(α::Real, β::Real, x::Real) = betalogcdf(promote(float(α), β, x)...)
-betalogcdf(k::T, x::T) where T = throw(MethodError(betalogcdf, (α, β, x)))
 
-function betalogccdf(α::Float64, β::Float64, x::Float64)
+function betalogccdf(α::Real, β::Real, x::Real)
     p, q = beta_inc(α, β, x)
     if q < 0.7
         return log(q)
@@ -58,13 +52,7 @@ function betalogccdf(α::Float64, β::Float64, x::Float64)
         return log1p(-p)
     end
 end
-betalogccdf(α::Real, β::Real, x::Real) = betalogccdf(promote(float(α), β, x)...)
-betalogccdf(α::T, β::T, x::T) where T = throw(MethodError(betalogccdf, (α, β, x)))
 
-betainvcdf(α::Float64, β::Float64, p::Float64) = first(beta_inc_inv(α, β, p, 1 - p))
-betainvcdf(α::Real, β::Real, p::Real) = betainvcdf(promote(float(α), β, p)...)
-betainvcdf(α::T, β::T, p::T) where T = throw(MethodError(betainvcdf, (α, β, p)))
+betainvcdf(α::Real, β::Real, p::Real) = first(beta_inc_inv(α, β, p, 1 - p))
 
-betainvccdf(α::Float64, β::Float64, p::Float64) = first(beta_inc_inv(α, β, 1 - p, p))
-betainvccdf(α::Real, β::Real, p::Real) = betainvccdf(promote(float(α), β, p)...)
-betainvccdf(α::T, β::T, p::T) where T = throw(MethodError(betainvccdf, (α, β, p)))
+betainvccdf(α::Real, β::Real, p::Real) = first(beta_inc_inv(α, β, 1 - p, p))
--- a/src/distrs/beta.jl
+++ b/src/distrs/beta.jl
@@ -1,8 +1,9 @@
 # functions related to beta distributions
 
 # R implementations
-# For pdf and logpdf we use the Julia implementation
 using .RFunctions:
+    # betapdf,
+    # betalogpdf,
     betacdf,
     betaccdf,
     betalogcdf,
--- a/src/distrs/beta.jl
+++ b/src/distrs/beta.jl
@@ -53,6 +53,6 @@ function betalogccdf(α::Real, β::Real, x::Real)
     end
 end
 
-betainvcdf(α::Real, β::Real, p::Real) = first(beta_inc_inv(α, β, p, 1 - p))
+betainvcdf(α::Real, β::Real, p::Real) = first(beta_inc_inv(α, β, p))
 
-betainvccdf(α::Real, β::Real, p::Real) = first(beta_inc_inv(α, β, 1 - p, p))
+betainvccdf(α::Real, β::Real, p::Real) = last(beta_inc_inv(α, β, p))
--- a/src/distrs/beta.jl
+++ b/src/distrs/beta.jl
@@ -26,17 +26,18 @@ function betalogpdf(α::T, β::T, x::T) where {T<:Real}
     return x < 0 || x > 1 ? oftype(val, -Inf) : val
 end
 
-betacdf(α::Real, β::Real, x::Real) = first(beta_inc(α, β, x))
+betacdf(α::Real, β::Real, x::Real) = first(beta_inc(α, β, min(max(0, x), 1)))
 
-betaccdf(α::Real, β::Real, x::Real) = last(beta_inc(α, β, x))
+betaccdf(α::Real, β::Real, x::Real) = last(beta_inc(α, β, min(max(0, x), 1)))
 
 # The log version is currently based on non-log version. When the cdf is very small we shift
 # to an implementation based on the hypergeometric function ₂F₁ to avoid underflow.
 function betalogcdf(α::T, β::T, x::T) where {T<:Real}
-    p, q = beta_inc(α, β, x)
+    _x = min(max(0, x), 1)
+    p, q = beta_inc(α, β, _x)
     if p < eps(one(p))
         # see https://dlmf.nist.gov/8.17#E7
-        return -log(α) + xlogy(α, x) + log(_₂F₁(promote(α, 1 - β, α + 1, x)...)) - logbeta(α, β)
+        return -log(α) + xlogy(α, _x) + log(_₂F₁(promote(α, 1 - β, α + 1, _x)...)) - logbeta(α, β)
     elseif p <= 0.7
         return log(p)
     else
@@ -46,7 +47,7 @@ end
 betalogcdf(α::Real, β::Real, x::Real) = betalogcdf(promote(α, β, x)...)
 
 function betalogccdf(α::Real, β::Real, x::Real)
-    p, q = beta_inc(α, β, x)
+    p, q = beta_inc(α, β, min(max(0, x), 1))
     if q < 0.7
         return log(q)
     else
--- a/src/distrs/beta.jl
+++ b/src/distrs/beta.jl
@@ -32,7 +32,7 @@ function betacdf(α::Real, β::Real, x::Real)
         return float(last(promote(α, β, x, x >= 0)))
     end
 
-    return first(beta_inc(α, β, min(max(0, x), 1)))
+    return first(beta_inc(α, β, clamp(x, 0, 1)))
 end
 
 function betaccdf(α::Real, β::Real, x::Real)
@@ -41,7 +41,7 @@ function betaccdf(α::Real, β::Real, x::Real)
         return float(last(promote(α, β, x, x < 0)))
     end
 
-    last(beta_inc(α, β, min(max(0, x), 1)))
+    last(beta_inc(α, β, clamp(x, 0, 1)))
 end
 
 # The log version is currently based on non-log version. When the cdf is very small we shift
@@ -52,7 +52,7 @@ function betalogcdf(α::T, β::T, x::T) where {T<:Real}
         return log(last(promote(α, β, x, x >= 0)))
     end
 
-    _x = min(max(0, x), 1)
+    _x = clamp(x, 0, 1)
     p, q = beta_inc(α, β, _x)
     if p < floatmin(p)
         # see https://dlmf.nist.gov/8.17#E7
@@ -71,7 +71,7 @@ function betalogccdf(α::Real, β::Real, x::Real)
         return log(last(promote(α, β, x, x < 0)))
     end
 
-    p, q = beta_inc(α, β, min(max(0, x), 1))
+    p, q = beta_inc(α, β, clamp(x, 0, 1))
     if q < 0.7
         return log(q)
     else
--- a/src/distrs/beta.jl
+++ b/src/distrs/beta.jl
@@ -26,13 +26,32 @@ function betalogpdf(α::T, β::T, x::T) where {T<:Real}
     return x < 0 || x > 1 ? oftype(val, -Inf) : val
 end
 
-betacdf(α::Real, β::Real, x::Real) = first(beta_inc(α, β, min(max(0, x), 1)))
+function betacdf(α::Real, β::Real, x::Real)
+    # Handle a degenerate case
+    if iszero(α)
+        return float(last(promote(α, β, x, x >= 0)))
+    end
 
-betaccdf(α::Real, β::Real, x::Real) = last(beta_inc(α, β, min(max(0, x), 1)))
+    return first(beta_inc(α, β, min(max(0, x), 1)))
+end
+
+function betaccdf(α::Real, β::Real, x::Real)
+    # Handle a degenerate case
+    if iszero(α)
+        return float(last(promote(α, β, x, x < 0)))
+    end
+
+    last(beta_inc(α, β, min(max(0, x), 1)))
+end
 
 # The log version is currently based on non-log version. When the cdf is very small we shift
 # to an implementation based on the hypergeometric function ₂F₁ to avoid underflow.
 function betalogcdf(α::T, β::T, x::T) where {T<:Real}
+    # Handle a degenerate case
+    if iszero(α)
+        return log(last(promote(α, β, x, x >= 0)))
+    end
+
     _x = min(max(0, x), 1)
     p, q = beta_inc(α, β, _x)
     if p < floatmin(p)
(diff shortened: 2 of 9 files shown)


andreasnoack:
I've added methods for Poisson via the definitions for gamma so this now supersedes #112. It requires https://github.com/JuliaRegistries/General/pull/37083 for the tests to pass so I'll restart once that PR is merged.

Project.toml
```diff
@@ -3,6 +3,7 @@ uuid = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 version = "0.9.8"
 
 [deps]
+HypergeometricFunctions = "34004b35-14d8-5ef3-9330-4cdb6864b03a"
```
devmotion, line 6: This requires a `[compat]` entry.

This also pulls in some additional dependencies since HypergeometricFunctions depends on DualNumbers, it seems? I am just wondering if people are fine with this since my ChainRules PR (https://github.com/JuliaStats/StatsFuns.jl/pull/106) was not received positively mainly due to the additional dependency on ChainRulesCore.

On a side note, HypergeometricFunctions would be needed in SpecialFunctions to resolve https://github.com/JuliaMath/SpecialFunctions.jl/issues/317 and https://github.com/JuliaMath/SpecialFunctions.jl/issues/160. There was some discussion about depending on HypergeometricFunctions in https://github.com/JuliaMath/SpecialFunctions.jl/pull/175, but I assume for the very same reasons not everyone would welcome such a dependency in SpecialFunctions.

src/distrs/beta.jl
```diff
@@ -1,19 +1,71 @@
...
-    betainvcdf,
-    betainvccdf,
+    # betapdf,
+    # betalogpdf,
+    # betacdf,
+    # betaccdf,
+    # betalogcdf,
+    # betalogccdf,
+    # betainvcdf,
+    # betainvccdf,
```
devmotion, line 13: I wonder why these lines are only commented out and not removed completely?

src/distrs/beta.jl
```diff
@@ -1,19 +1,71 @@
...
+betacdf(α::Float64, β::Float64, x::Float64) = first(beta_inc(α, β, x))
+betacdf(α::Real, β::Real, x::Real) = betacdf(promote(float(α), β, x)...)
+betacdf(k::T, x::T) where T = throw(MethodError(betacdf, (α, β, x)))
+
+betaccdf(α::Float64, β::Float64, x::Float64) = last(beta_inc(α, β, x))
+betaccdf(α::Real, β::Real, x::Real) = betaccdf(promote(float(α), β, x)...)
+betaccdf(k::T, x::T) where T = throw(MethodError(betaccdf, (α, β, x)))
+
+# The log version is currently based on non-log version. When the cdf is very small we shift
+# to a implementation based on the hypergeometric function ₂F₁ to avoid underflow.
```
devmotion, line 39: Is there a reference for these cutoffs?

src/distrs/beta.jl
```diff
@@ -1,19 +1,71 @@
...
     betainvlogccdf
 
 # pdf for numbers with generic types
-betapdf(α::Real, β::Real, x::Number) = x^(α - 1) * (1 - x)^(β - 1) / beta(α, β)
+betapdf(α::Real, β::Real, x::Real) = exp(betalogpdf(α, β, x))
 
 # logpdf for numbers with generic types
-betalogpdf(α::Real, β::Real, x::Number) = (α - 1) * log(x) + (β - 1) * log1p(-x) - logbeta(α, β)
+function betalogpdf(α::T, β::T, x::T) where T<:Real
+    return if x <= 0.5
```
devmotion, line 22: Maybe one could explain the motivation for why this is only used for ``x <= 0.5``?

src/distrs/beta.jl
```diff
@@ -1,19 +1,71 @@
...
+        xlogy(α - 1, x) + (β - 1) * log1p(-x) - logbeta(α, β)
+    else
+        betalogpdf(β, α, 1 - x)
+    end
+end
+betalogpdf(α::Real, β::Real, x::Real) = betalogpdf(promote(α, β, x)...)
+
+betacdf(α::Float64, β::Float64, x::Float64) = first(beta_inc(α, β, x))
+betacdf(α::Real, β::Real, x::Real) = betacdf(promote(float(α), β, x)...)
+betacdf(k::T, x::T) where T = throw(MethodError(betacdf, (α, β, x)))
```
devmotion, line 32: Is it actually necessary to add these `MethodErrors`? A `MethodError` would be returned anyways it seems? Also, is this supposed to be
```suggestion
betacdf(α::T, β::T, x::T) where T = throw(MethodError(betacdf, (α, β, x)))
```
?

src/distrs/beta.jl
```diff
@@ -1,19 +1,71 @@
...
+
+betaccdf(α::Float64, β::Float64, x::Float64) = last(beta_inc(α, β, x))
+betaccdf(α::Real, β::Real, x::Real) = betaccdf(promote(float(α), β, x)...)
+betaccdf(k::T, x::T) where T = throw(MethodError(betaccdf, (α, β, x)))
```
devmotion, line 36: Same here, is this needed and should this be
```suggestion
betaccdf(α::T, β::T, x::T) where T = throw(MethodError(betaccdf, (α, β, x)))
```
?

src/distrs/binom.jl
```diff
@@ -1,19 +1,25 @@
...
     binominvlogcdf,
     binominvlogccdf
 
-# pdf for numbers with generic types
-binompdf(n::Real, p::Real, k::Real) = exp(binomlogpdf(n, p, k))
+binompdf(n::Real, p::Real, k::Real) = betapdf(k + 1, n - k + 1, p)/(n + 1)
 
-# logpdf for numbers with generic types
-binomlogpdf(n::Real, p::Real, k::Real) = -log1p(n) - logbeta(n - k + 1, k + 1) + k * log(p) + (n - k) * log1p(-p)
+binomlogpdf(n::Real, p::Real, k::Real) = betalogpdf(k + 1, n - k + 1, p) - log(n + 1)
```
devmotion, line 17: I guess `log1p(n)` and is more expensive but not more accurate than `log(n + 1)` for `n >> 1`?

devmotion (commented):
Some initial comments but I'm too tired for a thorough review now. I'll check it more thoroughly tomorrow.

src/distrs/beta.jl
andreasnoack: This is to avoid a stack overflow for e.g. `Float32` elements.

src/distrs/binom.jl
andreasnoack: That was my thinking. `log1p` is only relevant for small arguments and that won't happen here.

Project.toml
andreasnoack: I'll add a compat entry (or CompatHelper will do it).

I can try to measure the effect on load time which I assume is the main concern here. I doubt that it will be much and more generally, I think that HypergeomericFunctions is a natural dependency here. Many CDFs have hypergeometric representations so I think this is a bit different from the AD discussion. I don't think it would make sense to have the CDFs defined in a separate package. (I think I'm in favor of #106 btw)

src/distrs/beta.jl
andreasnoack: I thought it was useful to keep them around while we are working on removing all of them. I'm also fine with deleting them, though.

src/distrs/beta.jl
devmotion: I see. Would it maybe be easier to just rely on the promotions in SpecialFunctions in https://github.com/JuliaMath/SpecialFunctions.jl/blob/fbb44faaa06f7b05723ac38811f7786d3142e0b5/src/beta_inc.jl#L880-L884 and call the four-argument version (unfortunately, it seems https://github.com/JuliaMath/SpecialFunctions.jl/blob/fbb44faaa06f7b05723ac38811f7786d3142e0b5/src/beta_inc.jl#L879 is only defined for `Float64` - I guess it should be generalized to `Real`?)?

src/distrs/beta.jl
andreasnoack: Unfortunately not. My thinking is that using the non-log version is fine as long as it doesn't underflow. I realize that I'm using `floatmin()` instead of `eps()` in a similar situation for one of the other distributions.

Project.toml
devmotion: I agree that probably it is more central to the package than AD (even though, of course, I'm in favor of the ChainRules PR :wink: ). I am fine with the dependency but I wanted to raise it due to the reactions in the other PR.

src/distrs/tdist.jl
```diff
@@ -13,7 +13,9 @@ import .RFunctions:
     tdistinvlogccdf
 
 # pdf for numbers with generic types
-tdistpdf(ν::Real, x::Number) = gamma((ν + 1) / 2) / (sqrt(ν * pi) * gamma(ν / 2)) * (1 + x^2 / ν)^(-(ν + 1) / 2)
+tdistpdf(ν::T, x::T) where T<:Real = gamma((ν + 1) / 2) / (sqrt(ν * π) * gamma(ν / 2)) * (1 + x^2 / ν)^(-(ν + 1) / 2)
+tdistpdf(ν::Real, x::Real) = tdistpdf(promote(ν, x)...)
 
 # logpdf for numbers with generic types
-tdistlogpdf(ν::Real, x::Number) = loggamma((ν + 1) / 2) - log(ν * pi) / 2 - loggamma(ν / 2) + (-(ν + 1) / 2) * log(1 + x^2 / ν)
+tdistlogpdf(ν::T, x::T) where T<:Real = loggamma((ν + 1) / 2) - log(ν * π) / 2 - loggamma(ν / 2) + (-(ν + 1) / 2) * log(1 + x^2 / ν)
```
devmotion, line 20: I guess here `log1p` could actually be useful:
```suggestion
tdistlogpdf(ν::T, x::T) where T<:Real = loggamma((ν + 1) / 2) - log(ν * π) / 2 - loggamma(ν / 2) + (-(ν + 1) / 2) * log1p(x^2 / ν)
```

src/distrs/beta.jl
```diff
@@ -1,19 +1,71 @@
...
+end
+betalogpdf(α::Real, β::Real, x::Real) = betalogpdf(promote(α, β, x)...)
+
+betacdf(α::Float64, β::Float64, x::Float64) = first(beta_inc(α, β, x))
+betacdf(α::Real, β::Real, x::Real) = betacdf(promote(float(α), β, x)...)
+betacdf(α::T, β::T, x::T) where T = throw(MethodError(betacdf, (α, β, x)))
+
+betaccdf(α::Float64, β::Float64, x::Float64) = last(beta_inc(α, β, x))
+betaccdf(α::Real, β::Real, x::Real) = betaccdf(promote(float(α), β, x)...)
+betaccdf(α::T, β::T, x::T) where T = throw(MethodError(betaccdf, (α, β, x)))
```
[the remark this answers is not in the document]
andreasnoack: I think we'd still need a promoting version somewhere. Either here or in SpecialFunctions.

src/distrs/beta.jl
```diff
@@ -1,19 +1,71 @@
...
+        # see https://dlmf.nist.gov/8.17#E7
+        return -log(α) + α*log(x) + log(_₂F₁(promote(α, 1 - β, α + 1, x)...)) - logbeta(α, β)
+    elseif p <= 0.7
+        return log(p)
+    else
+        return log1p(-q)
+    end
+end
+betalogcdf(α::Real, β::Real, x::Real) = betalogcdf(promote(float(α), β, x)...)
+betalogcdf(k::T, x::T) where T = throw(MethodError(betalogcdf, (α, β, x)))
```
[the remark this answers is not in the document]
andreasnoack: The idea was that the current implementations are restricted to `Float64` already and that we don't have generic versions available right now so that aspect should be handled separately anyway. I'd be okay with moving some of the promotion to SpecialFunctions though

src/distrs/beta.jl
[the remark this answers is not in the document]
devmotion: Sure, it's needed somewhere. To me it just seems it's better to have it in SpecialFunctions since it is not only of interest to StatsFuns. For the 4-argument version of `beta_inc`, these promotions (and some additional ones) are already contained in SpecialFunctions. I opened https://github.com/JuliaMath/SpecialFunctions.jl/pull/318 to extend them to the 3-argument version that is called here.

src/distrs/binom.jl
```diff
@@ -1,19 +1,25 @@
...
+
+binomcdf(n::Real, p::Real, k::Real) = betaccdf(k + 1, n - k, p)
+
+binomccdf(n::Real, p::Real, k::Real) = betacdf(k + 1, n - k, p)
+
+binomlogcdf(n::Real, p::Real, k::Real) = betalogccdf(k + 1, n - k, p)
+
+binomlogccdf(n::Real, p::Real, k::Real) = betalogcdf(k + 1, n - k, p)
```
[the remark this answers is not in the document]
andreasnoack: We'd then have to write the methods for the log versions again here and since the pattern is the same for the log and non-log versions, I think current version is preferable. If needed, I can add a comment.

src/distrs/fdist.jl
```diff
@@ -13,7 +13,11 @@ import .RFunctions:
...
-fdistpdf(ν1::Real, ν2::Real, x::Number) = sqrt((ν1 * x)^ν1 * ν2^ν2 / (ν1 * x + ν2)^(ν1 + ν2)) / (x * beta(ν1 / 2, ν2 / 2))
+fdistpdf(ν1::Real, ν2::Real, x::Real) = exp(fdistlogpdf(ν1, ν2, x))
 
 # logpdf for numbers with generic types
-fdistlogpdf(ν1::Real, ν2::Real, x::Number) = (ν1 * log(ν1 * x) + ν2 * log(ν2) - (ν1 + ν2) * log(ν1 * x + ν2)) / 2 - log(x) - logbeta(ν1 / 2, ν2 / 2)
+function fdistlogpdf(ν1::T, ν2::T, x::T) where T<:Real
+    ν1ν2 = ν1/ν2
+    return  xlogy(ν1/2, ν1ν2) + xlogy(ν1/2 - 1, x) - xlogy((ν1 + ν2)/2, 1 + ν1ν2*x) - logbeta(ν1/2, ν2/2)
+end
+fdistlogpdf(ν1::Real, ν2::Real, x::Real) = fdistlogpdf(promote(ν1, ν2, x)...)
```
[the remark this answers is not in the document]
andreasnoack: Yes. That should be doable.

src/distrs/beta.jl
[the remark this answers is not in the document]
andreasnoack: I missed that method when looking at your PR. Yes, let's just have the promotion there and simplify things here.

blackeneth:
Let me butt in here to comment. Andreas has functions (pdf cdf ccdf logpdf logccdf invcdf invccdf invlogcdf invlogccdf) for Beta, Binomial, Chi Squared, F dist, Gamma, Normal, Poisson, and Student's T distribution. 

A [long ago PR in Distributions.jl](https://github.com/JuliaStats/Distributions.jl/pull/138) has functions for Beta, Binomial, Cauchy, Chi Squared, F dist, Gamma, Geometric, HyperGeometric, logistic, lognormal, negative binomial, normal, poisson, student's t, uniform, weibull, and rayleigh distributions.

These pull requests always seem to stall. Review comments from devmotion tend to be along the lines of 1) small code optimizations, 2) questions on the "source" to certain algorithm constants. Multiply that by 8 for the all the distributions that Andreas is promoting. It results in a lot of little issues to fix/improve/modify. Review comments are valuable and it is leads to improved code, but one must manage the feedback process. 

Perhaps it would be better to do a separate pull request for each distribution. It is more manageable to fix/modify one function at a time, rather than a set of 8.  Learnings from fixing/modifying the first distribution would carry on to the second and subsequent distributions, so hopefully the process would be smoother. It is valuable to reference the code back to research papers; however, lacking that, we could depend on the test results matching the existing R functions. 

Also, one might consider a parallel release strategy.  Keep the R functions for now.  Release a pure Julia version in parallel -- this is the "Beta test" -- let the community use it, study it, and identify problems (such as corner cases).  After the Beta period, swap them for the R functions. 

A long term goal has been to eliminate the dependence on the R functions. Don't let the perfect be the enemy of the good. These functions need to make it over the finish line.

andreasnoack:
I think it's fine. @devmotion comments here are very valid and should be addressed. The main complication is that some of the changes require adjustments in SpecialFunctions which makes it slightly more work but it's not that much. I had to look into some other things for a while but I'll get this completed.

src/distrs/beta.jl
fernando-duarte: betalogpdf should return -Inf outside of (0,1)
```
    return if (x < zero(T) || x > one(T)) # is (x < 1 | x > 1) better?
        -Inf
    elseif x <= 0.5
```

fernando-duarte:
Tried to fix support for betalogpdf, cdf can use a similar fix I think 
https://github.com/JuliaStats/StatsFuns.jl/pull/113/files#r663370467

src/distrs/beta.jl
devmotion: This would not be type stable. Maybe one could just perform computations with `clamp(x, 0, 1)` and return `oftype(result, -Inf)` if `x` is outside the unit interval.

devmotion: Due to the check `x <= 0.5`, it would even be sufficient to perform computations with `max(0, x)`.

fernando-duarte: Why is it not type stable? `typeof(-Inf)` is `Float64`

fernando-duarte (approved):

src/distrs/beta.jl
fernando-duarte: Why is it not type stable? `typeof(-Inf)` is `Float64`
PS. I started a review by mistake, not sure how to remove myself from the reviewer list or dismiss my review -- sorry!

devmotion: > Why is it not type stable?

The other branches might not return a result of type `Float64` (e.g., if the inputs are of type `Float32`).

fernando-duarte: Good point. 

`-T(Inf)` then?

devmotion: No, this is incorrect and will error if the inputs are e.g. of type `Rational`.

fernando-duarte: Right, I am too narrowly focused on my own use case -- some day I will become less of a newb :) 

Thanks for the comments.

Pramodh-G:
Hello,
I stumbled across this issue while benchmarking R vs Distributions.jl sampling. I think having pure julia implementations can be really useful to the ecosystem as a whole. I want to help out with this issue wherever I can; Is there anything that I can do?

PS: I have little experience in writing numerical software but fluent in julia; given Rmath/similar implementations, I would be happy to port it to native julia :).

src/distrs/beta.jl
```diff
@@ -22,3 +25,35 @@ function betalogpdf(α::T, β::T, x::T) where {T<:Real}
...
+    if q < 0.7
+        return log(q)
+    else
+        return log1p(-p)
+    end
+end
+
+betainvcdf(α::Real, β::Real, p::Real) = first(beta_inc_inv(α, β, p))
+
+betainvccdf(α::Real, β::Real, p::Real) = last(beta_inc_inv(α, β, p))
```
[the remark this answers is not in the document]
devmotion: @andreasnoack I checked locally and this seems to fix the errors. The remaining errors are numerical differences between our and the R implementation for very tiny values. Interesting is e.g. that we are better in the following case which causes a test error:
```julia
x = 0.48
p = RFunctions.betacdf(1000, 2, x) # 9.0797754e-317 (same result as betacdf(1000, 2, x))
betainvcdf(1000, 2, p) # 0.4799999999640294
RFunctions.betainvcdf(1000, 2, p) # NaN
```

andreasnoack:
I've pushed a lot of fixes. It was great with all the tests for values out of the support as it revealed many places that required more care. On top of https://github.com/JuliaMath/SpecialFunctions.jl/pull/374 and https://github.com/JuliaMath/SpecialFunctions.jl/pull/376 I now have all tests passing locally.

src/distrs/beta.jl
```diff
@@ -22,3 +25,35 @@ function betalogpdf(α::T, β::T, x::T) where {T<:Real}
...
+        return log(p)
+    else
+        return log1p(-q)
+    end
+end
+betalogcdf(α::Real, β::Real, x::Real) = betalogcdf(promote(α, β, x)...)
+
+function betalogccdf(α::Real, β::Real, x::Real)
+    p, q = beta_inc(α, β, x)
+    if q < 0.7
```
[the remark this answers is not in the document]
andreasnoack: I couldn't find a way to write the complementary version in terms of the hypergeometric in the same was as I could with the logcdf.

src/distrs/gamma.jl
```diff
@@ -22,3 +25,58 @@ function gammalogpdf(k::T, θ::T, x::T) where {T<:Real}
     val = -loggamma(k) + xlogy(k - 1, xθ) - log(θ) - xθ
     return x < 0 ? oftype(val, -Inf) : val
 end
+
+gammacdf(k::T, θ::T, x::T) where {T<:Real} = first(gamma_inc(k, max(0, x)/θ))
+gammacdf(k::Real, θ::Real, x::Real)        = gammacdf(promote(float(k), θ, x)...)
```
devmotion, line 30: I think it would be better to promote first and convert to floating point numbers afterwards. Otherwise e.g. `gammacdf(1, 2f0, 5f0)` will return a `Float64`. So as below already for `gammalogcdf` maybe
```suggestion
gammacdf(k::Real, θ::Real, x::Real)        = gammacdf(map(float, promote(k, θ, x))...)
```

src/distrs/gamma.jl
```diff
@@ -22,3 +25,58 @@ function gammalogpdf(k::T, θ::T, x::T) where {T<:Real}
...
+
+gammaccdf(k::T, θ::T, x::T) where {T<:Real} = last(gamma_inc(k, max(0, x)/θ))
+gammaccdf(k::Real, θ::Real, x::Real)        = gammaccdf(promote(float(k), θ, x)...)
```
devmotion, line 33: Similarly
```suggestion
gammaccdf(k::Real, θ::Real, x::Real)        = gammaccdf(map(float, promote(k, θ, x))...)
```

src/distrs/gamma.jl
devmotion: But I wonder - is it actually necessary to convert to `float` here? Isn't this handled by `gamma_inc` in the line above?

src/distrs/gamma.jl
devmotion: Or again maybe just promote?

src/distrs/gamma.jl
```diff
@@ -22,3 +25,58 @@ function gammalogpdf(k::T, θ::T, x::T) where {T<:Real}
...
+
+gammalogcdf(k::Real, θ::Real, x::Real) = _gammalogcdf(map(float, promote(k, θ, x))...)
```
devmotion, line 35: Again, alternatively maybe split it into two definitions?
```suggestion
gammalogcdf(k::Real, θ::Real, x::Real) = gammalogcdf(promote(k, θ, x)...)
gammalogcdf(k::T, θ::T, x::T) where {T<:Real} = _gammalogcdf(float(k), float(θ), float(x))
```

src/distrs/gamma.jl
```diff
@@ -22,3 +25,58 @@ function gammalogpdf(k::T, θ::T, x::T) where {T<:Real}
...
+    end
+end
+
+function _gammalogccdf(k::T, θ::T, x::T) where {T<:Union{Float16,Float32}}
+    return T(_gammalogccdf(Float64(k), Float64(θ), Float64(x)))
+end
+
+function gammainvcdf(k::Real, θ::Real, p::Real)
+    _k, _θ, _p = map(float, promote(k, θ, p))
+    return _θ*gamma_inc_inv(_k, _p, 1 - _p)
```
devmotion, line 76: Does `gamma_inc_inv` not handle non-floating point arguments? Otherwise `map(float, ...)` might not be necessary.

andreasnoack:
Yay! All tests pass now and I believe I have addressed the review comments so I think this one is ready. It only took me about eight months.

src/distrs/beta.jl
```diff
@@ -22,3 +25,60 @@ function betalogpdf(α::T, β::T, x::T) where {T<:Real}
     val = xlogy(α - 1, y) + xlog1py(β - 1, -y) - logbeta(α, β)
     return x < 0 || x > 1 ? oftype(val, -Inf) : val
 end
+
+function betacdf(α::Real, β::Real, x::Real)
+    # Handle a degenerate case
+    if iszero(α)
+        return float(last(promote(α, β, x, x >= 0)))
```
devmotion, line 32: What's the motivation for this heuristic? And shouldn't the degenerate case maybe be handled in SpecialFunctions (IIRC you opened an issue)?

src/distrs/gamma.jl
```diff
@@ -22,3 +25,80 @@ function gammalogpdf(k::T, θ::T, x::T) where {T<:Real}
     val = -loggamma(k) + xlogy(k - 1, xθ) - log(θ) - xθ
     return x < 0 ? oftype(val, -Inf) : val
 end
+
+function gammacdf(k::T, θ::T, x::T) where {T<:Real}
+    # Handle the degenerate case
+    if iszero(k)
+        return last(promote(k, θ, x, x >= 0))/sqrt(one(θ))
```
devmotion, line 32: Similarly, what's the reason for this specific heuristic and should this maybe be handled in SpecialFunctions?

devmotion (commented):
Since the fixes in SpecialFunctions are necessary for the tests to pass, I guess it would be reasonable to update the SpecialFunctions compat entry (maybe the fixes should be backported to SpecialFunctions 1 if we want to keep support for SpecialFunctions 1.X)?

src/distrs/beta.jl
andreasnoack: When α is zero, the distribution is degenerate at zero so the CDF is just `x >= 0` but the return type would have to be adjusted.

I opened the issue in SpecialFunctions to discuss it there but I'm not completely sure that these edge cases belong in SpecialFunctions. Maybe they do but if you don't view the regularized incomplete beta function as a CDF then this limit might not be what you want. I assume the function might have applications outside of probability theory.

src/distrs/gamma.jl
andreasnoack: Same reasoning as above

andreasnoack:
I theory we should probably backport the SpecialFunctions PRs since I consider all of them bugfixes.

src/distrs/beta.jl
```diff
@@ -22,3 +25,60 @@ function betalogpdf(α::T, β::T, x::T) where {T<:Real}
...
+
+    last(beta_inc(α, β, clamp(x, 0, 1)))
+end
+
+# The log version is currently based on non-log version. When the cdf is very small we shift
+# to an implementation based on the hypergeometric function ₂F₁ to avoid underflow.
+function betalogcdf(α::T, β::T, x::T) where {T<:Real}
+    # Handle a degenerate case
+    if iszero(α)
+        return log(last(promote(α, β, x, x >= 0)))
```
devmotion, line 52:
```suggestion
        return log(last(promote(x, x >= 0)))
```

src/distrs/binom.jl
```diff
@@ -22,3 +22,27 @@ function binomlogpdf(n::T, p::T, k::T) where {T<:Real}
...
+
+        # When p == 1 the betaccdf doesn't return the correct result
+        # so these cases have to be special cased
+        if isone(p)
+            newres = oftype(res, $compl ? k < n : k >= n)
+            if $l === ""
+                return newres
+            else
+                return log(newres)
+            end
```
devmotion, line 43: Does this compile away? Or is it better to use

```suggestion
            return $(l === "" ? :newres : :(log(newres)))
```

src/distrs/gamma.jl
```diff
@@ -22,3 +25,80 @@ function gammalogpdf(k::T, θ::T, x::T) where {T<:Real}
...
+        return float(last(promote(x, x >= 0)))
+    end
+    return first(gamma_inc(k, max(0, x)/θ))
+end
+gammacdf(k::Real, θ::Real, x::Real) = gammacdf(map(float, promote(k, θ, x))...)
+
+function gammaccdf(k::T, θ::T, x::T) where {T<:Real}
+    # Handle the degenerate case
+    if iszero(k)
+        return last(promote(k, θ, x, x < 0))/sqrt(one(θ))
```
devmotion, line 41:
```suggestion
        return float(last(promote(x, x < 0)))
```

src/distrs/gamma.jl
```diff
@@ -22,3 +25,80 @@ function gammalogpdf(k::T, θ::T, x::T) where {T<:Real}
...
+gammaccdf(k::Real, θ::Real, x::Real) = gammaccdf(map(float, promote(k, θ, x))...)
+
+gammalogcdf(k::Real, θ::Real, x::Real) = _gammalogcdf(map(float, promote(k, θ, x))...)
+
+# Implemented via the non-log version. For tiny values, we recompute the result with
+# loggamma. In that situation, there is little risk of significant cancellation.
+function _gammalogcdf(k::Float64, θ::Float64, x::Float64)
+    # Handle the degenerate case
+    if iszero(k)
+        return log(last(promote(k, θ, x, x >= 0))/sqrt(one(θ)))
```
devmotion, line 54:
```suggestion
        return log(x >= 0)
```

src/distrs/gamma.jl
```diff
@@ -22,3 +25,80 @@ function gammalogpdf(k::T, θ::T, x::T) where {T<:Real}
...
+end
+
+gammalogccdf(k::Real, θ::Real, x::Real) = _gammalogccdf(map(float, promote(k, θ, x))...)
+
+# Implemented via the non-log version. For tiny values, we recompute the result with
+# loggamma. In that situation, there is little risk of significant cancellation.
+function _gammalogccdf(k::Float64, θ::Float64, x::Float64)
+    # Handle the degenerate case
+    if iszero(k)
+        return log(last(promote(k, θ, x, x < 0))/sqrt(one(θ)))
```
devmotion, line 78:
```suggestion
        return log(x < 0)
```

src/distrs/beta.jl
devmotion: Also in light of the discussion in SpecialFunctions, I agree that probably it's best to handle the special cases here but not in SpecialFunctions (as suggested also in https://github.com/JuliaMath/SpecialFunctions.jl/issues/378#issuecomment-1033369989).

Maybe the logic and the code would be slightly simpler here if we would focus on the special case `iszero(α) && !iszero(β) && iszero(x)` specifically (the other cases of `iszero(α)` should still be handled by SpecialFunctions, and in particular `iszero(α) && iszero(β)` should throw an error also here as it is not clear to which degenerate case it should correspond). I.e., one would only use
```julia
    # Not handled by SpecialFunctions but set to value of CDF of degenerate case here
    if iszero(α) && !iszero(β) && iszero(x)
        return float(one(first(promote(α, β, x))))
```

andreasnoack: There is also the question of the function value outside the unit interval which, again, is well defined for the CDF but not necessarily for the regularized incomplete beta function more generally so it might be simpler with the current version. The inequality definition is also the mathematical definition of CDF in the degenerate case so in some sense a very direct translation. Concretely, the branches based on the hypergeometric function also assume positive values of the parameters so I'd have to handle the zero cases outside there anyway.

andreasnoack: More generally, what I'm saying is that I don't see an issue with this branch covering a larger part than the parameter space than necessary. In particular because a single simple definition is sufficient. I realize, though, that I'll have to add a positivity condition to the non-zero parameter to avoid computing the wrong value in the case of `α = β = 0`.

I think it's fine to support the degenerate cases here even though we currently don't support them in Distributions. I suspect we might want to loosen the restriction in Distributions since it can be convenient e.g. during optimization. We loosened the requirement for the standard deviation in the Normal distribution a while back for the same reason.

devmotion: >  I realize, though, that I'll have to add a positivity condition to the non-zero parameter to avoid computing the wrong value in the case of α = β = 0.

Exactly, this was my main concern above 🙂

I'm fine with supporting the degenerate case, the condition in the `if` statement just seemed to cover too many other cases. But maybe it's simpler to only check `iszero(alpha) && !iszero(beta)` and cover all `x` instead of my suggestion above.

andreasnoack: I've updated the code.

Amusingly, I realized that
```
julia> StatsFuns.RFunctions.betacdf(0.0, 0.0, 0.0)
0.0

julia> StatsFuns.RFunctions.betacdf(0.0, 0.0, 1e-100)
0.5

julia> StatsFuns.RFunctions.betacdf(0.0, 0.0, 0.9999)
0.5

julia> StatsFuns.RFunctions.betacdf(0.0, 0.0, 1.0)
1.0
```
but I'm pretty sure that we shouldn't support the Bernoulli limit since the success probability depends on the ratio between `α` and `β` as they go to zero.

devmotion: Yes, I agree.

andreasnoack:
@devmotion The fixes have been backported in https://github.com/JuliaRegistries/General/pull/54891 so are we good to go here?

devmotion (approved):
Looks good to me! Great PR 🎉

PR #118, opened 2021-07-10
Depend on IrrationalConstants.jl

torfjelde:
Now that [IrrationalConstants.jl](https://github.com/JuliaMath/IrrationalConstants.jl) have been released, the constants in StatsFuns are no longer required.

This is nice since it means that we can take advantange the constants that before were defined in StatsFuns.jl in other packages, without depending on StatsFuns.jl, e.g. some of them would be useful in SpecialFunctions.jl.

EDIT: This has also been done in LogExpFunctions.jl :+1:


src/misc.jl
```diff
@@ -7,7 +7,7 @@ Return the logarithm of [multivariate gamma function](https://en.wikipedia.org/w
 """
 function logmvgamma(p::Int, a::Real)
```
torfjelde, line 8: Test for this, i.e. `src/misc.jl:45`, is currently failing because the numerical accuracy is _slightly_ different than `logbeta`. I'm very confused as to why it started failing now as the constants aren't even being used in `logmvgamma`.

This came up before I made the change to use `logπ` btw; that was an attempt to fix it.

devmotion:
This will lead to ambiguity warnings and redefinitions with LogExpFunctions 0.2, won't it? It seems a bit more consistent also to update LogExpFunctions to 0.3 which depends on IrrationalConstants. Similar to LogExpFunctions maybe one should not reexport the constants anymore and require users to load IrrationalConstants explicitly, @tpapp?

torfjelde:
> This will lead to ambiguity warnings and redefinitions with LogExpFunctions 0.2, won't it? It seems a bit more consistent also to update LogExpFunctions to 0.3 which depends on IrrationalConstants.

Ah yes sorry, checked that LogExpFunctions now depends on IrrationalConstants.jl but forgot to bump the compat version :+1:

torfjelde:
Also should prob wait for https://github.com/JuliaMath/SpecialFunctions.jl/pull/333 since otherwise bumping LogExpFunctions.jl to 0.3 will lead to a downgrade of SpecialFunctions.

devmotion:
SpecialFunctions 1.6.1 supports LogExpFunctions 0.3.

torfjelde:
Aaaaah you're back my love :heart: 

Should we bump the compat entry for SpecialFunctions then?

torfjelde:
No I mean, this PR will be breaking if you can't use IrrationalConstants.jl, right? So if IrrationalConstants.jl is only compat with SpecialFunctions 1.6.1, then should we lower bound the compat entry for SpecialFunctiosn here in StatsFuns for to 1.6.1?

torfjelde:
Haha, of course; nevermind me! :upside_down_face: 

Sooo are we good then, with the exception of that one test which now seems to fail because of "numerical improvement"?

And what should I do with that test? Just relax the rtol a bit?

devmotion:
I've seen these test failures before, maybe they are caused by some changes in SpecialFunctions. Can it be reproduced on the master branch?

Otherwise, I think we still have to address https://github.com/JuliaStats/StatsFuns.jl/pull/118#issuecomment-877657834 - currently both LogExpFunctions (0.2) and IrrationalConstants define and export these constants.

devmotion:
We could also support both LogExpFunctions 0.2 and 0.3, it seems, by changing `using LogExpFunctions` to an explicit `using LogExpFunctions: logit, ...` without the constants.

devmotion:
Fixed and extended by https://github.com/JuliaStats/StatsFuns.jl/pull/122.

PR #119, opened 2021-08-11
Make CompatHelper run on v1

ChrisRackauckas:


DilumAluthge:
I'd recommend you use the default Julia that ships on GHA runners. You can copy-paste the recommended config from here: https://github.com/JuliaRegistries/CompatHelper.jl/blob/master/.github/workflows/CompatHelper.yml

PR #120, opened 2021-08-11
Update LogExpFunctions compat

ChrisRackauckas:


devmotion:
In LogExpFunctions 0.3 the irrational constants are not exported anymore. StatsFuns should depend on IrrationalConstants explicitly. https://github.com/JuliaStats/StatsFuns.jl/pull/118 should fix this.

PR #123, opened 2021-08-15
Fix CompatHelper.yml

devmotion:
Turns out the official CompatHelper workflow in CompatHelper.jl contained a bug (https://github.com/JuliaRegistries/CompatHelper.jl/pull/376). It seems everything will become easier soon with compathelper-action (https://github.com/JuliaRegistries/CompatHelper.jl/pull/379).

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -1,4 +1,3 @@
-
 name: CompatHelper
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


devmotion:
I ran the CompatHelper action manually with this PR to ensure that it works properly now: https://github.com/JuliaStats/StatsFuns.jl/runs/3332641954?check_suite_focus=true

PR #124, opened 2021-09-03
Allow arguments of type `Float16` and `Float32` in R functions

devmotion:
Currently it is not possible to call R functions such as `nchisqlogpdf` with arguments of type `Float16` or `Float32` which makes it impossible to use these types in e.g. `NoncentralChisq` in Distributions (https://github.com/JuliaStats/Distributions.jl/pull/1387#discussion_r700418572). This PR fixes the problem.


src/rmath.jl
```diff
@@ -6,6 +6,8 @@ import Rmath: libRmath
 
 ### import macro
 
+const FloatOrInt = Union{Float16,Float32,Float64,Int}
```
nalimilan, line 9: Any reason not to accept smaller integer types? Maybe we could even allow any `Real`, and let the conversion throw in case of overflow?

devmotion: I did not know where to stop so I just started with the types that would fix the linked issue in Distributions :smile: I was wondering about smaller integer types as well. I think `Real` would break stuff and lead to method ambiguities since for some of these functions we already define Julia implementations that accept `Real` arguments in StatsFuns (they are used e.g. for `Dual` numbers).

nalimilan: Hmm, then why not use the Julia implementations when they exist? Is it faster to call Rmath even if that involves conversions?

devmotion: This is the motivation for https://github.com/JuliaStats/StatsFuns.jl/pull/113 which I did not want to touch or duplicate in this PR. I haven't done any benchmarks but in my opinion performance should not be the main criterion for switching to Julia implementations. For instance, we should make sure that the Julia implementation is at least as accurate as the Rmath version, ideally on the whole input domain.

nalimilan: Yeah #113 is complex enough that it's better to leave it alone. I'm just wondering what's the criterion to add some types to the list but not others. Maybe at least add small integer types?

devmotion: I added `IntX`s, `UIntX`s, and `Rational`s to the union as well since it would fix e.g. issues such as https://github.com/JuliaStats/Distributions.jl/pull/1387#issuecomment-913115634. I am not particularly happy about the hardcoded union and hence used `Real` instead whenever no Julia fallback existed.

src/rmath.jl
```diff
@@ -109,6 +140,9 @@ macro import_rmath(rname, jname, pargs...)
     esc(_import_rmath(rname, jname, pargs))
 end
 
+macro import_rmath_real(rname, jname, pargs...)
```
nalimilan, line 143:
```suggestion
# Used when no Julia fallback exists currently
macro import_rmath_real(rname, jname, pargs...)
```

src/rmath.jl
nalimilan: OK, thanks.  Some `Rational{Int}` values cannot be converted to `Float64` without a loss in precision, but maybe we don't care, just like `promote_type(Int64, Float64) == Float64` despite the precision loss for large values?

It would be nice to test the new types. :-)

devmotion:
Replaced and fixed by https://github.com/JuliaStats/StatsFuns.jl/pull/125.

PR #125, opened 2021-09-06
Allow `Real` arguments in R functions and use Julia implementations only

devmotion:
This PR is based on #124 and contains some parts of @andreasnoack's PR #113. I thought it would be cleaner to open a new PR instead of adding commits to #124.

The main motivation for this PR is that I am not satisfied with the approach in #124. I agree with @nalimilan that the type union is somewhat arbitrary and ideally we would just allow `Real` and let `convert(Float64, ...)` (implicitly in the `ccall`) handle the arguments when we call into Rmath. The main issue is that currently some generic Julia implementations exist in StatsFuns that already allow `Real` but they are used only as fallbacks and not for arguments of type `Float64` or `Int`.

In this PR,
- R functions are defined for arguments of type `Real`,
- and the Julia implementations do not extend the corresponding function in the `RFunctions` submodule anymore but are separate functions of the same name to which also `Float64` and `Int` are dispatched.

The second point makes it also a bit easier to ensure that we actually compare the R with the Julia implementations in the tests since currently e.g. `RFunctions.betalogpdf` and `betalogpdf` are the same function and for arguments of type `Union{Float64,Int}` always the R implementation is used.

Currently, the Julia implementations do not correspond to the R versions exactly, e.g. `betapdf` errors for values of `x` outside of `[0, 1]` whereas `RFunctions.betapdf` returns 0. I used some of @andreasnoack's changes in #113 and tried to handle values outside of the support in this PR. I also extended the tests similar to #113, in addition to the tests with Float16 and Float32 from #124, to make sure that the Julia implementations are accurate enough. Maybe some additional tests should be added though.

**The main difference from #113 is that this PR does not add any new Julia implementations (e.g. for CDF or other distributions). It only updates the existing implementations such that they can be used instead of the R functions also with `Float64` and `Int`.**

A positive side-effect of using the Julia implementation for `Float64` is that it seems to improve performance. The example below is motivated by @cscherrer's observation (https://twitter.com/ChadScherrer/status/1433143576943337477) that MeasureTheory's pdf for student t distributions is faster than `tdistpdf`.

### Current release

```julia
julia> using StatsFuns, BenchmarkTools, Random

julia> N = 100_000;

julia> out = Vector{Float64}(undef, N);

julia> @btime map!(betapdf, $out, $(randexp(N)), $(randexp(N)), $(rand(N)));
  32.881 ms (0 allocations: 0 bytes)

julia> @btime map!(betalogpdf, $out, $(randexp(N)), $(randexp(N)), $(rand(N)));
  29.748 ms (0 allocations: 0 bytes)

julia> @btime map!(tdistpdf, $out, $(randexp(N)), $(randn(N)));
  35.008 ms (0 allocations: 0 bytes)

julia> @btime map!(tdistlogpdf, $out, $(randexp(N)), $(randn(N)));
  31.259 ms (0 allocations: 0 bytes)
```

### This PR

```julia
julia> using StatsFuns, BenchmarkTools, Random

julia> N = 100_000;

julia> out = Vector{Float64}(undef, N);

julia> @btime map!(betapdf, $out, $(randexp(N)), $(randexp(N)), $(rand(N)));
  10.250 ms (0 allocations: 0 bytes)

julia> @btime map!(betalogpdf, $out, $(randexp(N)), $(randexp(N)), $(rand(N)));
  9.373 ms (0 allocations: 0 bytes)

julia> @btime map!(tdistpdf, $out, $(randexp(N)), $(randn(N)));
  7.560 ms (0 allocations: 0 bytes)

julia> @btime map!(tdistlogpdf, $out, $(randexp(N)), $(randn(N)));
  6.860 ms (0 allocations: 0 bytes)
```


src/distrs/pois.jl
```diff
@@ -12,19 +12,8 @@ import .RFunctions:
...
     poisinvlogccdf
 
-# generic versions
+# Julia implementations
 poispdf(λ::Real, x::Real) = exp(poislogpdf(λ, x))
 
+poislogpdf(λ::Real, x::Real) = poislogpdf(promote(λ, x)...)
 poislogpdf(λ::T, x::T) where {T <: Real} = xlogy(x, λ) - λ - loggamma(x + 1)
-
-poislogpdf(λ::Number, x::Number) = poislogpdf(promote(float(λ), x)...)
```
nalimilan, line 20: Do you know why `float` was used previously?

src/chainrules.jl
```diff
@@ -12,9 +12,9 @@ ChainRulesCore.@scalar_rule(
     binomlogpdf(n::Real, p::Real, k::Real),
     @setup(z = digamma(n - k + 1)),
     (
-        digamma(n + 2) - z + log1p(-p) - 1 / (1 + n),
+        ChainRulesCore.NoTangent(),
```
nalimilan, line 15: I have no idea what this does. Is it needed for this PR or could it be applied separately?

test/rmath.jl
```diff
@@ -116,26 +119,53 @@ end
...
-        ((10, 2), 0.01:0.01:0.99),
+        ((0.1, 1.0), 0.0:0.01:1.0),
+        ((1.0, 1.0), 0.0:0.01:1.0),
+        ((2.0, 3.0), 0.0:0.01:1.0),
+        ((10.0, 2.0), 0.0:0.01:1.0),
+        ((10, 2), 0.0:0.01:1.0),
+        ((1f0, 1f0), 0f0:0.01f0:1f0),
+        ((1.0, 1.0), 0f0:0.01f0:1f0),
+        ((Float16(1), Float16(1)), Float16(0):Float16(0.01):Float16(1)),
+        ((1f0, 1f0), Float16(0):Float16(0.01):Float16(1)),
```
nalimilan, line 130: I'd also test other types such as integers, rationals and/or a big type. Also, do we have tests somewhere mixing different input types to check that promotion works?

nalimilan (commented):
Sounds reasonable. I guess that switching progressively to Julia implementations can also make the process easier than waiting for a complete switch. I'll leave @andreasnoack comment on these implementations though.

test/rmath.jl
devmotion: Yes, I'll add integers and rationals as well. Promotions are already tested, e.g. here Float32 and Float16 and Float32 and Float64 above.

src/chainrules.jl
devmotion: It is required since otherwise tests fail: currently we test derivatives with respect to these arguments with a `BigFloat` hack - only the generic Julia implementation allows to perturb the integer-valued arguments but the R implementation is not defined (or rather returns -Inf) for these values. In this PR the Julia implementations are updated to match this behaviour (which e.g. is also expected by Distributions) and returns `-Inf` if `n` or `k` are no integers. So we can't test these derivatives anymore. And due to this definition of `binomlogpdf` it does not make sense to define them based on the continuous extension anymore (I think it's questionable if it ever did).

src/distrs/pois.jl
devmotion: No.

devmotion: Maybe to ensure that `poislogpdf(::Rational{Int}, ::Int)` also hits the R fallback?

test/rmath.jl
devmotion: I added more tests with integers and rationals. I also added tests for the values outside of the support.

devmotion:
Bump :slightly_smiling_face: 

 It would also be nice to merge this PR since it fixes the remaining test errors in https://github.com/JuliaStats/Distributions.jl/issues/1387.

devmotion:
@andreasnoack it would be great if you could have a look at this PR :slightly_smiling_face:

andreasnoack (approved):

PR #128, opened 2021-10-26
Cancel Github actions for previous commits in PRs

devmotion:
This PR updates the Github actions setup and cancels runs for all but the latest commit in each PR automatically. I think this is particularly useful for the more time consuming integration tests. Additionally, coverage upload issues don't affect the status of the jobs anymore.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
       - uses: julia-actions/julia-runtest@v1
       - uses: julia-actions/julia-processcoverage@v1
         continue-on-error: true
-      - uses: codecov/codecov-action@v1
+      - uses: codecov/codecov-action@v2
         continue-on-error: true
         with:
           file: lcov.info
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,5 @@
 name: CI
 
-env:
-  JULIA_NUM_THREADS: 2
-
 on:
   pull_request:
     branches:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -43,6 +52,8 @@ jobs:
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
       - uses: julia-actions/julia-processcoverage@v1
+        continue-on-error: true
```
nalimilan, line 55: What's the advantage of this? I'd rather be warned when coverage reporting failed.

devmotion: As mentioned in https://github.com/JuliaMath/SpecialFunctions.jl/pull/357#discussion_r735590713, it can be helpful to let the coverage upload etc. not influence the Github test status. Lately, I also had some problems with coverage uploads since the coverage services were down and it is annoying if this causes test failures. One could still make the coverage checks mandatory on Github to prevent merging PRs with significant coverage drops.

I can revert the change if not desired though, I just used the same settings as in SpecialFunctions.

nalimilan: I find it useful to get a big red icon when coverage failed for some reason. That gives you a hint that something might be wrong. You can still ignore it if it turns out to be wrong, but sometimes it catches real lacks in coverage.

devmotion: I don't mind, so I'll just revert these changes.

devmotion: Done.

devmotion:
I addressed your comment @nalimilan, are you satisfied with the PR or would you like some additional changes?

nalimilan:
Sorry, I had forgotten this!

PR #132, opened 2022-01-22
Promote arguments of `norminvcdf` and `norminvccdf`

devmotion:
Currently these functions compute `erfcinv` in possibly lower precision and only take into account the parameters (and their types) of the normal distribution afterwards. This PR fixes and tests this problem.

### DIFF
--- a/src/distrs/norm.jl
+++ b/src/distrs/norm.jl
@@ -91,14 +91,14 @@ end
 norminvcdf(p::Real) = -erfcinv(2*p) * sqrt2
 function norminvcdf(μ::Real, σ::Real, p::Real)
     # Promote to ensure that we don't compute erfcinv in low precision and then promote
-    _μ, _σ, _p = map(float, promote(μ, σ, p))
+    _μ, _σ, _p = promote(μ, σ, p)
     xval(_μ, _σ, norminvcdf(_p))
 end
 
 norminvccdf(p::Real) = erfcinv(2*p) * sqrt2
 function norminvccdf(μ::Real, σ::Real, p::Real)
     # Promote to ensure that we don't compute erfcinv in low precision and then promote
-    _μ, _σ, _p = map(float, promote(μ, σ, p))
+    _μ, _σ, _p = promote(μ, σ, p)
     xval(_μ, _σ, norminvccdf(_p))
 end
 
--- a/src/distrs/norm.jl
+++ b/src/distrs/norm.jl
@@ -89,18 +89,14 @@ function normlogccdf(μ::Real, σ::Real, x::Number)
 end
 
 norminvcdf(p::Real) = -erfcinv(2*p) * sqrt2
-function norminvcdf(μ::Real, σ::Real, p::Real)
-    # Promote to ensure that we don't compute erfcinv in low precision and then promote
-    _μ, _σ, _p = promote(μ, σ, p)
-    xval(_μ, _σ, norminvcdf(_p))
-end
+# Promote to ensure that we don't compute erfcinv in low precision and then promote
+norminvcdf(μ::Real, σ::Real, p::Real) = norminvcdf(promote(μ, σ, p)...)
+norminvcdf(μ::T, σ::T, p::T) where {T<:Real} = xval(μ, σ, norminvcdf(p))
 
 norminvccdf(p::Real) = erfcinv(2*p) * sqrt2
-function norminvccdf(μ::Real, σ::Real, p::Real)
-    # Promote to ensure that we don't compute erfcinv in low precision and then promote
-    _μ, _σ, _p = promote(μ, σ, p)
-    xval(_μ, _σ, norminvccdf(_p))
-end
+# Promote to ensure that we don't compute erfcinv in low precision and then promote
+norminvccdf(μ::Real, σ::Real, p::Real) = norminvccdf(promote(μ, σ, p)...)
+norminvccdf(μ::T, σ::T, p::T) where {T<:Real} = xval(μ, σ, norminvccdf(p))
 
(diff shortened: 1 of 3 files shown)


src/distrs/norm.jl
```diff
@@ -82,17 +82,25 @@ normlogccdf(z::Number) = z > 1.0 ?
...
+        z = zval(μ, σ, x)
     end
     normlogccdf(z)
 end
 
 norminvcdf(p::Real) = -erfcinv(2*p) * sqrt2
-norminvcdf(μ::Real, σ::Real, p::Real) = xval(μ, σ, norminvcdf(p))
+function norminvcdf(μ::Real, σ::Real, p::Real)
+    # Promote to ensure that we don't compute erfcinv in low precision and then promote
+    _μ, _σ, _p = map(float, promote(μ, σ, p))
```
sethaxen, line 94: Couldn't you just:
```suggestion
    _μ, _σ, _p = promote(μ, σ, float(p))
```
?

src/distrs/norm.jl
```diff
@@ -82,17 +82,25 @@ normlogccdf(z::Number) = z > 1.0 ?
...
+    xval(_μ, _σ, norminvcdf(_p))
+end
 
 norminvccdf(p::Real) = erfcinv(2*p) * sqrt2
-norminvccdf(μ::Real, σ::Real, p::Real) = xval(μ, σ, norminvccdf(p))
+function norminvccdf(μ::Real, σ::Real, p::Real)
+    # Promote to ensure that we don't compute erfcinv in low precision and then promote
+    _μ, _σ, _p = map(float, promote(μ, σ, p))
```
sethaxen, line 101: Same here
```suggestion
    _μ, _σ, _p = promote(μ, σ, float(p))
```

test/rmath.jl
```diff
@@ -17,7 +17,18 @@ end
...
 get_rmathfun(fname) = eval(Meta.parse(string("RFunctions.", fname)))
 
-function rmathcomp(basename, params, X::AbstractArray, rtol=100eps(float(one(eltype(X)))))
+function rmathcomp(basename, params, X::AbstractArray, rtol=nothing)
+    # compute desired tolerance:
+    # has to take into account `params` as well since otherwise e.g. `X::Array{<:Rational}`
+    # always uses a tolerance based on `eps(one(Float64))` even when parameters are of type
+    # Float32
+    rtol = if rtol === nothing
+        PT = mapreduce(typeof, promote_type, params)
```
sethaxen, line 26: Couldn't you use one of the promotion function in Base here?

test/rmath.jl
```diff
@@ -17,7 +17,18 @@ end
...
+        100 * eps(float(one(promote_type(PT, eltype(X)))))
+    else
+        rtol
+    end
```
sethaxen, line 30:
```suggestion
    if rtol === nothing
        PT = mapreduce(typeof, promote_type, params)
        rtol = 100 * eps(float(one(promote_type(PT, eltype(X)))))
    end
```

src/distrs/norm.jl
devmotion: No, this will cause undesired promotions: If e.g. `mu` and `sigma` are of type `Float32` and `p` is of type `Int` or `Rational` (more likely, eg. in Distributions fallback for `median` uses `1//2`), then the PR would promote to and return `Float32` whereas your suggestion would promote to and return `Float64`.

sethaxen: Ah, good point.

devmotion: The `float` could be removed though, I just copied it from Distributions but SpecialFunctions calls it already: https://github.com/JuliaMath/SpecialFunctions.jl/blob/414385a7097e0f376b1273a3e0deac9d30cf0f9e/src/erf.jl#L397

test/rmath.jl
devmotion: Is there one that promotes `(1, 1//2)` or `(1f0, 2.0)` to `Rational{Int}` and `Float64`?

test/rmath.jl
devmotion: I prefer the version in the PR since in my opinon it makes it easier to reason about the value of `rtol` and is easier to read.

src/distrs/norm.jl
devmotion: I simplified the promotion.

test/rmath.jl
sethaxen:
```julia
julia> Base.promote_typeof(1, 1//2)
Rational{Int64}

julia> Base.promote_typeof(1f0, 2.0)
Float64
```

```suggestion
        PT = Base.promote_typeof(params...)
```

devmotion: I had in mind that splatting in such a case is much slower (generally). And indeed, e.g.,
```julia
julia> using BenchmarkTools

julia> f(x) = Base.promote_typeof(x...)
f (generic function with 1 method)

julia> g(x) = mapreduce(typeof, promote_type, x)
g (generic function with 1 method)

julia> x = ntuple(_ -> (c = rand(); c < 0.2 ? rand(Float64) : (c < 0.7 ? rand(Int) : rand(Int) // rand(Int))), 100);

julia> @btime f($x);
  94.985 μs (101 allocations: 3.14 KiB)

julia> @btime g($x);
  18.967 μs (101 allocations: 3.14 KiB)
```
However, I guess it does not matter here since (a) it's only for tests and (b) the number of parameters, i.e., length of the tuples, is very small (I think <= 3 at most). So I'll use your suggestion to improve readability of the code :slightly_smiling_face:

nalimilan (approved):

Issue #136, opened 2022-04-01
Using gamma_inc method not defined in SpecialFunctions minimum version


sethaxen:
This package sets the lower bound for SpecialFunctions compatibility to v0.8.0. However, it uses the 2-arg `gamma_inc` method, which was not defined until SpecialFunctions v1.2.0. Since the Julia lower bound for SpecialFunctions v1.2.0 was v1.3, this results in `gammacdf` erroring for all pre-v1.3 versions of Julia, even though the Julia lower bound on this package is v1.0. 

This could be resolved by using the 3-arg method of `gamma_inc`. It would also be good to add the minimum supported Julia version to the CI.

devmotion:
I don't think it's useful to support such old versions of SpecialFunctions at all. There are multiple numerical issues without the recent bugfixes in SpecialFunctions 1.x and 2.x (we backported them to 1.x). So probably it is more appropriate to update the Julia compat.

sethaxen:
Since SpecialFunctions v1.2 does not support pre-Julia v1.3 versions at all, simply bumping the SpecialFunctions compat leaves older Julia versions with a version of this package for which `gammacdf` errors, right? That needs to be fixed first.

rikhuijzer:
> Since SpecialFunctions v1.2 does not support pre-Julia v1.3 versions at all, simply bumping the SpecialFunctions compat leaves older Julia versions with a version of this package for which `gammacdf` errors, right? That needs to be fixed first.

Possibly, but how much effort do we want to put into supporting Julia pre 1.6? Julia 1.6 was released more than a year ago.

sethaxen:
I did not suggest we should support older Julia versions. The most recent release of this package introduced the 2-arg `gamma_inc` method, which was a bug for versions of this package that it claims to support. I am merely suggesting we fix that bug before dropping support for those versions, or we are just leaving them with a bug.

Otherwise we coerce downstream dependencies to either all drop support for pre-Julia v1.3 so their CI passes or stop testing their minimum supported Julia version, which is not great.

devmotion:
I understand but unfortunately the package will still not work correctly on these older Julia versions. If you use the latest Rmath-reduced versions you will run into numerical issues and errors (reported in StatsFuns and eg in Distributions) if you don't use the latest releases of SpecialFunctions 1 or 2.

devmotion:
[...]

PR #137, opened 2022-04-01
Update CI

rikhuijzer:
- Test minimum supported version (came up in #136).
- Use `julia-actions/cache`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.0'
+          - '1.6'
           - '1' # Latest stable release.
         os:
           - ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,7 @@ on:
     branches:
       - master
     tags: '*'
+  workflow_dispatch:
 
 concurrency:
   # Skip intermediate builds: always.
@@ -26,7 +27,8 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1'
+          - '1.0'
+          - '1' # Latest stable release.
         os:
           - ubuntu-latest
           - macos-latest
@@ -39,16 +41,7 @@ jobs:
         with:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
-      - uses: actions/cache@v1
-        env:
-          cache-name: cache-artifacts
-        with:
-          path: ~/.julia/artifacts
-          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
-          restore-keys: |
-            ${{ runner.os }}-test-${{ env.cache-name }}-
-            ${{ runner.os }}-test-
-            ${{ runner.os }}-
+      - uses: julia-actions/cache@v1
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
       - uses: julia-actions/julia-processcoverage@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -39,16 +41,7 @@ jobs:
...
-        env:
-          cache-name: cache-artifacts
-        with:
-          path: ~/.julia/artifacts
-          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
-          restore-keys: |
-            ${{ runner.os }}-test-${{ env.cache-name }}-
-            ${{ runner.os }}-test-
-            ${{ runner.os }}-
+      - uses: julia-actions/cache@v1
```
devmotion, line 44: Is it safe to cache packages here? I had many issues with packages (such as e.g. PyCall) that mutate the package directory in platform-specific ways, and often had to disable it.

devmotion:
We require recent bug fixes in SpecialFunctions v1 and 2, so IMO it would be more honest to bump the Julia compat: https://github.com/JuliaStats/StatsFuns.jl/issues/136#issuecomment-1085529331

.github/workflows/ci.yml
rikhuijzer: Did you have issues outside of PyCall? Normally, the artifacts system should be very robust with caching. It has never caused me any problems.

devmotion: No, not artifacts but packages are the problem. And yes, I think other packages, or their dependencies, were problematic as well but I don't remember which.

Project.toml
```diff
@@ -21,7 +21,7 @@ LogExpFunctions = "0.3.2"
 Reexport = "1"
 Rmath = "0.4, 0.5, 0.6, 0.7"
 SpecialFunctions = "0.8, 0.9, 0.10, 1.0, 2"
-julia = "1"
+julia = "1.6"
```
devmotion, line 24: Unfortunately, we can't do this in a patch release, the registry doesn't allow it.

rikhuijzer: Can you finish this PR?

.github/workflows/ci.yml
rikhuijzer: Should that be an issue at `julia-actions/cache`? I think so. `cache-packages` can be set to false if you prefer that

rikhuijzer: Were your issues all related to Conda? See https://github.com/julia-actions/cache/issues/10 for a discussion

devmotion: The problem is also any other package that saves something to the package directory which is not platform independent. So yeah I think only caching artifacts would be a safer default since the errors where quite hard to debug.

rikhuijzer: It's a problem that also occurs for people manually moving depot paths over in HPC environments and that kind of stuff. Hiding the problem silently at the cache action is not a long-term solution.

devmotion: Hiding? Just setting the default to `cache_packages=false` would be much safer IMO. One could still opt in, but there is no way to rule out these issues as long as packages write to their package directory, e.g., to cache stuff, store paths, build stuff etc.

Project.toml
devmotion: Waiting on feedback on the general issue: https://github.com/JuliaStats/StatsFuns.jl/issues/136#issuecomment-1085700322

devmotion:
I checked everything more carefully again, and it seems sufficient to bump the Julia compat to 1.3. I'm waiting on more general feedback on the issue as well: https://github.com/JuliaStats/StatsFuns.jl/issues/136#issuecomment-1085700322

devmotion:
We will fix this differently: https://github.com/JuliaStats/StatsFuns.jl/issues/136#issuecomment-1085960412 Thanks for the PR nevertheless!

PR #138, opened 2022-04-01
Fix Julia 1.0 compatibility

### LINKED ISSUE #136: Using gamma_inc method not defined in SpecialFunctions minimum version

devmotion:
Fix Julia 1.0 compatibility by reverting https://github.com/JuliaStats/StatsFuns.jl/pull/113 and fixing some test tolerances.

I don't think there is a good way to fix Julia 1.0 compatibility without reverting that PR since only SpecialFunctions 0.8 supports Julia 1.0 but many fixes in more recent SpecialFunctions versions are required. E.g., `gamma_inc_inv` contains an undefined variable `r` in SpecialFunctions 0.8, i.e., fixing this problem would require us to reimplement `gamma_inc_inv` in StatsFuns.

#113 will be reapplied and released as part of StatsFuns 0.10.0 for Julia >= 1.3 and recent versions of SpecialFunctions.

Fixes #136.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
+          - '1.0'
           - '1'
         os:
           - ubuntu-latest
--- a/Project.toml
+++ b/Project.toml
@@ -4,7 +4,6 @@ version = "0.9.16"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
-HypergeometricFunctions = "34004b35-14d8-5ef3-9330-4cdb6864b03a"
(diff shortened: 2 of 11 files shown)


devmotion:
@nalimilan This bugfix and the follow-up PR are ready for review 🙂

PR #139, opened 2022-04-01
Reapply #113, update compat entries, and prepare StatsFuns 1.0.0

devmotion:
This PR is a follow-up to #138 and reapplies #113. It updates the compat entries and tests the oldest supported Julia version + LTS to ensure that users don't end up with (major) regressions compared with StatsFuns 0.9.17 that uses Rmath instead of Julia implementations of these functions.

An additional question is if we want to stop reexporting IrrationalConstants and LogExpFunctions. Since we make a breaking release this would be a good opportunity if it is desired.

Edit: I removed the reexports.


devmotion:
Hmm seems I messed up and lost the desired changes along the way.

devmotion:
Fixed by https://github.com/JuliaStats/StatsFuns.jl/pull/142.

PR #141, opened 2022-04-02
Consistent return type from `lstirling_asym(x::Float32)`

### LINKED ISSUE #140: return type of `lstirling_asym(x::Float32)` is inconsistent.
At present the return type for `x <= 40` is `Float64` whereas for `x > 40` it is `Float32`.
```jl
julia> StatsFuns.lstirling_asym(21.f0)
0.00396795421864086

julia> StatsFuns.lstirling_asym(41.f0)
0.00203248f0
```

I believe that the definition of lstirlingF32 should be changed to
```
const lstirlingF32 = Float32[lstirling_asym(k) for k in big(1):big(40)]
```
for consistency.

dmbates:
As described in #140.  I also changed the construction to use the previously calculated Float64 values.

Closes #140


src/misc.jl
```diff
@@ -66,7 +66,7 @@ lstirling_asym(x::BigFloat) = loggamma(x) + x - log(x)*(x - big(0.5)) - log2π/b
 lstirling_asym(x::Integer) = lstirling_asym(float(x))
 
 const lstirlingF64 = Float64[lstirling_asym(k) for k in big(1):big(64)]
-const lstirlingF32 = Float64[lstirling_asym(k) for k in big(1):big(40)]
+const lstirlingF32 = Float32.(view(lstirlingF64, 1:40))
```
devmotion, line 69: We could avoid the broadcasting machinery here and use something simpler (for the compiler at least) such as
```suggestion
const lstirlingF32 = [Float32(lstirlingF64[i]) for i in 1:40]
```

dmbates:
Thanks for the suggestion @devmotion

devmotion (commented):
LGTM. Can you add a test of the type (in)stability? Eg `@inferred StatsFuns.lstirling_asym(x)` where `x isa Float32`.

dmbates:
I hope this is good to go now.

devmotion (approved):

PR #142, opened 2022-04-20
Reapply #113 and prepare release of StatsFuns 1.0.0

devmotion:
This PR is a cleaned and corrected version of #139. Copied from there:

> This PR is a follow-up to https://github.com/JuliaStats/StatsFuns.jl/pull/138 and reapplies https://github.com/JuliaStats/StatsFuns.jl/pull/113. It updates the compat entries and tests the oldest supported Julia version + LTS to ensure that users don't end up with (major) regressions compared with StatsFuns 0.9.17 that uses Rmath instead of Julia implementations of these functions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,8 +26,9 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.0'
-          - '1'
+          - '1.3' # oldest supported version
+          - '1.6' # LTS
+          - '1' # latest release
         os:
           - ubuntu-latest
           - macos-latest
--- a/Project.toml
+++ b/Project.toml
@@ -19,9 +19,9 @@ InverseFunctions = "0.1"
 IrrationalConstants = "0.1"
 LogExpFunctions = "0.3.2"
 Reexport = "1"
-Rmath = "0.4, 0.5, 0.6, 0.7"
-SpecialFunctions = "0.8, 0.9, 0.10, 1.0, 2"
-julia = "1"
+Rmath = "0.6.1, 0.7"
+SpecialFunctions = "1.8.4, 2.1.4"
+julia = "1.3"
 
 [extras]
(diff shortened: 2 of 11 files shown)


devmotion:
@nalimilan I did not remove the reexports for now since it seems there was no general agreement in favour of this change (at least I remember you preferred them to be kept).

nalimilan:
> @nalimilan I did not remove the reexports for now since it seems there was no general agreement in favour of this change (at least I remember you preferred them to be kept).

Yeah my point was that if we drop IrrationalConstants and LogExpFunctions reexports, this package only contains distributions-related functions, so it would better be renamed e.g. DistributionsFuns. Maybe at some point we could split out these functions to a new package, and StatsFuns would become a meta package and/or be deprecated.

Before releasing 1.0, are we sure we are happy with all of the API? AFAICT it's been very stable so I guess it's fine?

andreasnoack (approved):

devmotion:
> Before releasing 1.0, are we sure we are happy with all of the API? AFAICT it's been very stable so I guess it's fine?

Yes, in my opinion it has been quite stable, I think the main consideration was just whether to drop the reexports or not.

In any case, we can always move to 2.0 if something comes up that is worth changing but breaking.

andreasnoack:
> In any case, we can always move to 2.0 if something comes up that is worth changing but breaking.

If that is our thinking then there is there is little reason to release 1.0. IMO releasing 1.0 only makes sense if we expect to commit to the API for some time.

nalimilan:
It doesn't seem there are any problems with the API (and my request on Slack didn't prompt any comments), so I'd say it's fine to release 1.0 -- unless you think we should drop reexports.

devmotion:
I don't mind either way, I think it's OK to keep them (for now at least). Just a final check as a follow-up to our discussion above, are you happy with the PR @andreasnoack?

andreasnoack:
Yes. Let's move forward with this as it is.

PR #148, opened 2022-11-10
Test source with JET

andreasnoack:
This revealed some bugs in the untested bivariate T cdf. I've fixed the bugs and also added a few tests.

### DIFF
--- a/src/tvpack.jl
+++ b/src/tvpack.jl
@@ -12,7 +12,7 @@
 # Original source available from
 # http://www.math.wsu.edu/faculty/genz/software/fort77/tvpack.f
 
-function tvtcdf(nu::Int, h::Vector{Float64}, r::Vector{Float64})
+function tvtcdf(nu::Int, h::NTuple{3,Float64}, r::NTuple{3,Float64})
     pt = 0.5 * pi
     h1 = h[1]
     h2 = h[2]
@@ -128,10 +128,10 @@ end
 # # One Dimensional Globally Adaptive Integration Function
 
 function adonet(f::Function, a::Float64, b::Float64, tol::Float64)
-    ai = Vector{Float64}(100)
-    bi = Vector{Float64}(100)
-    ei = Vector{Float64}(100)
-    fi = Vector{Float64}(100)
+    ai = Vector{Float64}(undef, 100)
+    bi = Vector{Float64}(undef, 100)
+    ei = Vector{Float64}(undef, 100)
+    fi = Vector{Float64}(undef, 100)
     ai[1] = a
     bi[1] = b
     err = 1.0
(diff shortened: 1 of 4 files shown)


test/runtests.jl
```diff
@@ -5,3 +5,8 @@ for t in tests
     println("* running $fp")
     include(fp)
 end
+
+using JET
+@testset "Static analysis" begin
+    @test isempty(JET.get_reports(report_package("StatsFuns", target_modules=(StatsFuns,))))
+end
```
devmotion, line 12: I think this should be checked only on Julia >= 1.7 (see also the comment in the logs: https://github.com/JuliaStats/StatsFuns.jl/actions/runs/3436051370/jobs/5729130463#step:6:88):
```suggestion

if VERSION >= v"1.7"
    using JET
    @testset "Static analysis" begin
        @test isempty(JET.get_reports(report_package("StatsFuns", target_modules=(StatsFuns,))))
    end
end
```

src/tvpack.jl
```diff
@@ -242,7 +242,9 @@ end
 
 function bvtcdf(nu::Int, dh::Float64, dk::Float64, r::Float64)
     if nu < 1
-        return bvnuppercdf(-dh, -dk, r)
+        # The source call the commented out line below. Instead we error out
+        throw(DomainError(nu, "degrees of freedom parameter must be positive"))
+        # return bvnuppercdf(-dh, -dk, r)
```
devmotion, line 247: That is, in tvpack.f this branch is supported? And we error since `bvnuppercdf` does not exist/was not copied?

andreasnoack: We do have `bvnuppercdf` but I think it's weird behavior to call it when the degrees of freedom parameter is non-positive.

devmotion: I see and I agree, it seems strange to support non-positive degrees. Maybe in tvpack it is just a way to encode the limit case of infinte degrees of freedom?

In any case, shouldn't `tcdf` be changed as well for consistency? That supports non-positive degrees of freedom as well it seems.

andreasnoack: It should. I hadn't noticed that. Will push an update.

andreasnoack: Reading this again. The branch is for `nu < 1`. I'd read it as `< 0`. It doesn't makes sense at all to use the normal cdf when the degrees of freedom parameter is between zero and one. Not sure what the idea here was.

devmotion:
It seems this PR is almost ready? Only https://github.com/JuliaStats/StatsFuns.jl/pull/148/files#r1019101826 is missing, it seems?

andreasnoack:
I'm not completely sure what to do with the issue mentioned in https://github.com/JuliaStats/StatsFuns.jl/pull/148#discussion_r1030585347

test/runtests.jl
```diff
@@ -5,3 +5,10 @@ for t in tests
     println("* running $fp")
     include(fp)
 end
+
+if VERSION >= v"1.7"
+    using JET
+    @testset "Static analysis" begin
+        @test isempty(JET.get_reports(report_package("StatsFuns", target_modules=(StatsFuns,))))
```
devmotion, line 12: One could also use `JET.test_package`, e.g.,
```suggestion
        JET.test_package("StatsFuns", target_defined_modules=true)
```

devmotion: Or
```suggestion
        JET.test_package("StatsFuns", target_defined_modules=true)
```

PR #149, opened 2022-11-24
Implement the T-distribution via the F-distribution

andreasnoack:
Today I realized that

1. I already implemented a pure Julia version of the F-distribution functions
2. We can just implement the `T(ν)`-distribution via the `F(1,ν)`-distribution 

Hence this PR. This is an alternative to #147

### DIFF
--- a/src/distrs/tdist.jl
+++ b/src/distrs/tdist.jl
@@ -42,7 +42,7 @@ function tdistinvcdf(ν::T, p::T) where T<:Real
     elseif p < 0.5
         return -sqrt(fdistinvccdf(one(ν), ν, 2*p))
     else
-        sqrt(fdistinvccdf(one(ν), ν, 2*(1 - p)))
+        return sqrt(fdistinvccdf(one(ν), ν, 2*(1 - p)))
     end
 end
 tdistinvcdf(ν::Real, p::Real) = tdistinvcdf(map(float, promote(ν, p))...)
--- a/src/distrs/tdist.jl
+++ b/src/distrs/tdist.jl
@@ -1,19 +1,5 @@
 # functions related to student's T distribution
 
-# R implementations
-using .RFunctions:
-    # tdistpdf,
-    # tdistlogpdf,
-    tdistcdf,
-    tdistccdf,
-    tdistlogcdf,
-    tdistlogccdf,
-    tdistinvcdf,
-    tdistinvccdf,
-    tdistinvlogcdf,
-    tdistinvlogccdf
-
-# Julia implementations
 tdistpdf(ν::Real, x::Real) = exp(tdistlogpdf(ν, x))
 
 tdistlogpdf(ν::Real, x::Real) = tdistlogpdf(promote(ν, x)...)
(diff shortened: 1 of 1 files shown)


src/distrs/tdist.jl
```diff
@@ -22,3 +8,67 @@ function tdistlogpdf(ν::T, x::T) where {T<:Real}
...
+tdistcdf(ν::Real, x::Real) = tdistcdf(map(float, promote(ν, x))...)
+
+tdistccdf(ν::Real, x::Real) = tdistcdf(ν, -x)
+
+function tdistlogcdf(ν::T, x::T) where T<:Real
+    if isinf(ν)
+        return normlogcdf(x)
+    elseif x < 0
+        ret = fdistlogccdf(one(ν), ν, x*x)
+        return ret - log(2*one(ret))
```
ararslan, line 30: FWIW I believe this should also work as expected and avoids needing to actually compute the log:
```suggestion
        return ret - logtwo
```

src/distrs/tdist.jl
```diff
@@ -22,3 +8,67 @@ function tdistlogpdf(ν::T, x::T) where {T<:Real}
     νp12 = (ν + 1) / 2
     return loggamma(νp12) - (logπ + log(ν)) / 2 - loggamma(ν / 2) - νp12 * log1p(x^2 / ν)
 end
+
+function tdistcdf(ν::T, x::T) where T<:Real
+    if isinf(ν)
+        return normcdf(x)
+    elseif x < 0
+        return fdistccdf(one(ν), ν, x*x)/2
```
devmotion, line 16: I would assume nowadays the compiler is smart enough to compile optimized code if we use
```suggestion
        return fdistccdf(one(ν), ν, x^2)/2
```
instead?

andreasnoack:
Thanks. I've pushed a commit that addresses the comments.

src/distrs/tdist.jl
```diff
@@ -22,3 +8,67 @@ function tdistlogpdf(ν::T, x::T) where {T<:Real}
...
+
+tdistlogccdf(ν::Real, x::Real) = tdistlogcdf(ν, -x)
+
+function tdistinvcdf(ν::T, p::T) where T<:Real
+    if isinf(ν)
+        return norminvcdf(p)
+    elseif p < 0.5
+        return -sqrt(fdistinvccdf(one(ν), ν, 2*p))
+    else
+        sqrt(fdistinvccdf(one(ν), ν, 2*(1 - p)))
```
ararslan, line 45:
```suggestion
        return sqrt(fdistinvccdf(one(ν), ν, 2*(1 - p)))
```

src/distrs/tdist.jl
```diff
@@ -22,3 +8,67 @@ function tdistlogpdf(ν::T, x::T) where {T<:Real}
...
+    end
+end
+tdistinvcdf(ν::Real, p::Real) = tdistinvcdf(map(float, promote(ν, p))...)
+
+tdistinvccdf(ν::Real, p::Real) = -tdistinvcdf(ν, p)
+
+if VERSION < v"1.7.0-beta1"
```
ararslan, line 52:
```suggestion
if VERSION < v"1.7.0-DEV.1172"
```
As an aside, perhaps this method should be added to Compat.

src/distrs/tdist.jl
```diff
@@ -22,3 +8,67 @@ function tdistlogpdf(ν::T, x::T) where {T<:Real}
...
+end
+_expm1(x::Number) = expm1(x)
+
+function tdistinvlogcdf(ν::T, logp::T) where T<:Real
+    if isinf(ν)
+        return norminvlogcdf(logp)
+    elseif logp < -log(2)
+        return -sqrt(fdistinvlogccdf(one(ν), ν, logp + logtwo))
+    else
+        sqrt(fdistinvccdf(one(ν), ν, -2*_expm1(logp)))
```
ararslan, line 69:
```suggestion
        return sqrt(fdistinvccdf(one(ν), ν, -2*_expm1(logp)))
```

ararslan (approved):
Very minor comments around consistency but otherwise looks great. Math checks out AFAICT and the tests are happy. Thanks!

Issue #150, opened 2022-12-02
gammalogcdf(k, θ, x) returns error for valid arguments


the-noble-argon:
I'm getting some errors in my code for some edge cases of the gammalogcdf function:

```julia
k = 42648.50647826826
θ = 2.2498007956420723e-5
x = 0.6991377135675367
gammalogcdf(k, θ, x)
```
which returns
```julia
ERROR: DomainError with -1.1090323347913147:
log will only return a complex result if called with a complex argument. Try log(Complex(x)).
Stacktrace:
 [1] throw_complex_domainerror(f::Symbol, x::Float64)
   @ Base.Math .\math.jl:33
 [2] _log(x::Float64, base::Val{:ℯ}, func::Symbol)
   @ Base.Math .\special\log.jl:301
 [3] log
   @ .\special\log.jl:267 [inlined]
 [4] _gammalogcdf(k::Float64, θ::Float64, x::Float64)
   @ StatsFuns C:\Users\user\.julia\packages\StatsFuns\0IeNQ\src\distrs\gamma.jl:60
 [5] gammalogcdf(k::Float64, θ::Float64, x::Float64)
   @ StatsFuns C:\Users\user\.julia\packages\StatsFuns\0IeNQ\src\distrs\gamma.jl:47
 [6] top-level scope
   @ c:\Users\user\Documents\julia_debug.jl:7
```
Internally this is because of the `log(drummond1F1(1.0, 1 + k, xdθ))` expression on line 60 of gamma.jl. For this input we get 
```julia
k = 42648.50647826826
xdθ = 31075.538550870202
drummond1F1(1.0, 1 + k, xdθ)
-1.1090323347913147
```
which is a negative number that you can't take the log of.

the-noble-argon:
I know that for this case, `k` is absolutely huge. Maybe you can case out the situation where k is really big and so the gamma distribution is essentially Normal?

PR #151, opened 2022-12-02
Use _₁F₁ instead of drummond1F1 in gammalogcdf

andreasnoack:
After the native Julia implementation for the Gamma distribution was implemented, HypergeometricFunctions added the _₁F₁ function which will decide on the algorithm based on the arguments. This ensures good behavior in a broader part of the parameter space.

Fixes https://github.com/JuliaStats/StatsFuns.jl/issues/150

### DIFF
--- a/src/distrs/gamma.jl
+++ b/src/distrs/gamma.jl
@@ -1,6 +1,6 @@
 # functions related to gamma distribution
 
-using HypergeometricFunctions: drummond1F1
+using HypergeometricFunctions: _₁F₁
 
 # R implementations
 using .RFunctions:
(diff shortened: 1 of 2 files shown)


src/distrs/gamma.jl
```diff
@@ -1,6 +1,6 @@
 # functions related to gamma distribution
 
-using HypergeometricFunctions: drummond1F1
+using HypergeometricFunctions: _₁F₁
```
devmotion, line 3: I assume we might have to update the compat entry for HypergeometricFunctions?

andreasnoack: It looks like it was introduced in 0.3.7, https://github.com/JuliaMath/HypergeometricFunctions.jl/commit/251d4221b7d6dccbe00f554948ac582299cee10d, so the current bound should be sufficient.

devmotion (approved):

PR #155, opened 2023-01-31
Remove code for unsupported Julia versions

devmotion:
When looking through tdist.jl, I noticed that there is some code left for nowadays unsupported Julia versions.

Edit: Somehow read 0.7 instead of 1.7, so clearly one of the branches is still needed...

Edit2: Maybe it would actually be better to use `log2mexp` here to continue performing operations in log-space.


devmotion:
Hmm, I think `log2mexp` would be a better approach - but it seems it requires adding a `_expm1(::Float64)` definition to LogExpFunctions?

devmotion:
The remaining test errors are caused by the recent Rmath update, see https://github.com/JuliaStats/StatsFuns.jl/pull/156

andreasnoack (approved):

PR #157, opened 2023-01-31
Fix tests

devmotion:


test/generic.jl
```diff
@@ -5,34 +5,38 @@ using ForwardDiff: Dual
...
 
-function genericcomp(basename, params, X::AbstractArray, rtol=100eps(float(one(eltype(X)))))
+function genericcomp(basename, params, X::AbstractArray)
+    # compute default tolerance:
+    # has to take into account `params` as well since otherwise e.g. `X::Array{<:Rational}`
+    # always uses a tolerance based on `eps(one(Float64))` even when parameters are of type
+    # Float32
+    # eps^(7//8) means requiring equality of about 7/8 of the significant digits
+    # Corresponds to tolerances of ~2e-14 (Float64), ~9f-7 (Float32) and ~0.002 (Float16)
+    rtol = eps(float(one(promote_type(Base.promote_typeof(params...), eltype(X)))))^(7//8)
```
devmotion, line 24: This is even a tiny bit smaller than the current tolerance (currently `100eps(Float64)  = 2.220446049250313e-14` whereas `eps(Float64)^(7//8) = 2.0097183471152322e-14`). But exponentiation seems to be more useful/interpretable generally, in particular when working with other floating point types. So I think we should change the default tolerance in rmath.jl as well, since there we actually work with other types such as `Float16` and `Float32`.

test/utils.jl
```diff
@@ -0,0 +1,14 @@
...
+    # Float32
+    return _default_rtol(float(promote_type(Base.promote_typeof(params...), eltype(X))))
+end
+
+# We use less sharp tolerances for Float16 and Float32 since the proportion of significant
+# digits that are equal when evaluating Rmath and StatsFuns is smaller
+# eps^(7//8) means requiring equality of about 7/8 of the significant digits, etc.
+_default_rtol(::Type{Float64}) = eps(Float64)^(7//8) # ~2.0e-14
+_default_rtol(::Type{Float32}) = eps(Float32)^(3//4) # ~6.4e-6
+_default_rtol(::Type{Float16}) = eps(Float16)^(2//3) # ~9.9e-3
```
devmotion, line 14: These tolerances are smaller than the tolerances on the master branch but arguably more interpretable:
```julia
julia> eps(Float64)^(7//8)
2.0097183471152322e-14

julia> 100eps(Float64)
2.220446049250313e-14

julia> eps(Float32)^(3//4)
6.4155306f-6

julia> 100eps(Float32)
1.1920929f-5

julia> eps(Float16)^(2//3)
Float16(0.00986)

julia> 100eps(Float16)
Float16(0.09766)
```

PR #168, opened 2024-09-02
Update GH actions and add dependabot

devmotion:
The PR updates the version of the GH actions (which allows us to use eg `min` and `lts` to specify Julia versions), runs tests on the latest pre-release instead of nightly (supposedly to give more relevant and stable results), and adds a dependabot config.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,10 +26,10 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.3' # oldest supported version
-          - '1.6' # LTS
+          - min # oldest supported version
+          - lts # LTS
           - '1' # latest release
-          - 'nightly'
+          - pre # pre-releases
         os:
           - ubuntu-latest
           - macos-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -26,35 +26,28 @@ jobs:
...
           - ubuntu-latest
           - macos-latest
           - windows-latest
         arch:
           - x64
     steps:
-      - uses: actions/checkout@v2
-      - uses: julia-actions/setup-julia@v1
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@v2.4.0
```
andreasnoack, line 41: Any specific reason to include minor and patch part here?

devmotion: In 2.4.0 support for the `min` version specifier was added. But the patch part is not needed, so probably we rather want
```suggestion
      - uses: julia-actions/setup-julia@v2.4
```

andreasnoack (approved):

PR #169, opened 2024-09-06
add more inverses

aplavin:
I personally just need logit/logistic, but others also make sense to add


devmotion:
These functions are defined in LogExpFunctions, not in StatsFuns. There the inverses are defined as well: https://github.com/JuliaStats/LogExpFunctions.jl/blob/289114f535827c612ce10c01b8dec9d3a55e4d15/ext/LogExpFunctionsInverseFunctionsExt.jl#L6-L14

PR #173, opened 2025-02-19
add signrank distribution

ArnoStrouwen:
Implementation of the signrank distribution.
The `cdf` is heavily optimized, since that is what is needed for the popular hypothesis test associated with this distribution.

The testing in this package seems quite involved, I could use some pointers on where to add the tests.

```jl
using StatsFuns
using Rmath

signrankpdf.(-2:12,4)
dsignrank.(-2:12,4,false)

signranklogpdf.(-2:12,4)
dsignrank.(-2:12,4,true)

signrankcdf.(-2:12,4)
psignrank.(-2:12,4,true,false)

signranklogcdf.(-2:12,4)
psignrank.(-2:12,4,true,true)

signrankccdf.(-2:12,4)
psignrank.(-2:12,4,false,false)

signranklogccdf.(-2:12,4)
psignrank.(-2:12,4,false,true)

signrankinvcdf.(-0.01:0.01:1.01,4)
qsignrank.(-0.01:0.01:1.01,4,true,false)
signrankinvcdf.(0.0624,4)
signrankinvcdf.(0.0625,4)
signrankinvcdf.(0.0626,4)
qsignrank.(0.0624,4,true,false)
qsignrank.(0.0625,4,true,false)
qsignrank.(0.0626,4,true,false)

signrankinvlogcdf.(log.(0.0:0.01:1.01),4)
qsignrank.(log.(0.0:0.01:1.01),4,true,true) # 0.0 does not match, is the R definition correct?
qsignrank.(log(0.0),4,true,true)
qsignrank.(0.0,4,true,false) 
signrankinvlogcdf.(log(0.0624),4)
signrankinvlogcdf.(log(0.0625),4)
signrankinvlogcdf.(log(0.0626),4)
qsignrank.(log(0.0624),4,true,true)
qsignrank.(log(0.0625),4,true,true)
qsignrank.(log(0.0626),4,true,true)


signrankinvccdf.(-0.01:0.01:1.01,4)
qsignrank.(-0.01:0.01:1.01,4,false,false) 
signrankinvccdf.(0.0624,4)
signrankinvccdf.(0.0625,4)
signrankinvccdf.(0.0626,4)
qsignrank.(0.0624,4,false,false)
qsignrank.(0.0625,4,false,false)
qsignrank.(0.0626,4,false,false)

signrankinvlogccdf.(log.(0.0:0.01:1.01),4)
qsignrank.(log.(0.0:0.01:1.01),4,false,true) # 0.0 does not match
signrankinvlogccdf.(log(0.0624),4)
signrankinvlogccdf.(log(0.0625),4)
signrankinvlogccdf.(log(0.0626),4)
qsignrank.(log(0.0624),4,false,true)
qsignrank.(log(0.0625),4,false,true)
qsignrank.(log(0.0626),4,false,true)

```

### DIFF
--- a/src/distrs/signrank.jl
+++ b/src/distrs/signrank.jl
@@ -1,0 +1,109 @@
+#=
+The signrank distribution is equivalent to the problem
+of finding the number of subsets of {1,2,...,n} summing to W,
+relative to the total number of subsets.
+The empty subset is defined to sum to zero.
+This can be calculated using the recursion:
+either n is in the subset in which case we need to calculate
+the number of subsets of {1,2,...,n-1} summing to W-n,
+or n is not in the subset in which case we need to calculate
+the number of subsets of {1,2,...,n-1} summing to W.
+This can be calculated bottom up using dynamic programming.
+
+The i'th element of DP in the j'th outer loop iteration represents:
+the number of ways {1,2,...,j} can sum to W-i+1.
+ =#
+
+@inline function signrankDP(n, W)
+    DP = zeros(Int, W + 1)
+    DP[W+1] = 1
+    for j in 1:n
+        for i in 1:(W+1-j)
+            DP[i] += DP[i+j]
+        end
+    end
+    return DP
+end
+
+function signrankpdf(n::Int, W::Union{Float64,Int})
+    f, _ = modf(W)
+    if f != 0
+        return 0.0
+    end
+    W = Int(W)
+    max_W = (n * (n + 1)) >> 1
+    if W < 0
+        return 0.0
+    elseif W > max_W >> 1
+        return signrankpdf(n, max_W - W)
+    end
+    DP = signrankDP(n, W)
+    return ldexp(float(DP[1]), -n)
+end
+
+function signranklogpdf(n::Int, W::Union{Float64,Int})
+    return log(signrankpdf(n, W))
+end
+
+function signrankcdf(n::Int, W::Union{Float64,Int})
+    W = round(Int, W, RoundNearestTiesUp)
+    max_W = (n * (n + 1)) >> 1
+    if W < 0
+        return 0.0
+    elseif W > max_W >> 1
+        return 1.0 - signrankcdf(n, max_W - W - 1)
+    end
+    DP = signrankDP(n, W)
+    return sum(Base.Fix2(ldexp, -n) ∘ float, DP)
+end
+
+function signranklogcdf(n::Int, W::Union{Float64,Int})
+    return log(signrankcdf(n, W))
+end
+
+function signrankccdf(n::Int, W::Union{Float64,Int})
+    max_W = (n * (n + 1)) >> 1
+    W = round(Int, W, RoundNearestTiesUp)
+    return signrankcdf(n, max_W - W - 1,)
+end
+
+function signranklogccdf(n::Int, W::Union{Float64,Int})
+    return log(signrankccdf(n, W))
+end
+
+function signrankinvcdf(n::Int, p::Float64)
+    if p < 0.0 || p > 1.0
+        return NaN
+    end
+    W = 0
+    while signrankcdf(n, W) < p # TODO binary search and symmetry
+        W += 1
+    end
+    return float(W)
+end
+
+function signrankinvlogcdf(n::Int, logp::Float64)
+    if logp > 0.0 || logp == -Inf
+        return NaN
+    end
+    W = 0
+    while signranklogcdf(n, W) < logp # TODO binary search and symmetry
+        W += 1
+    end
+    return float(W)
+end
+
+function signrankinvccdf(n::Int, p::Float64)
+    signrankinvcdf(n, 1 - p)
+end
+
+function signrankinvlogccdf(n::Int, logp::Float64)
+    if logp == -Inf
+        return NaN
+    end
+    if logp == 0
+        return 0.0
+    end
+    #signrankinvlogcdf(n, log1mexp(logp)) # does not roundtrip well
+    signrankinvccdf(n, exp(logp))
+end
--- a/src/distrs/signrank.jl
+++ b/src/distrs/signrank.jl
@@ -25,17 +25,17 @@ the number of ways {1,2,...,j} can sum to W-i+1.
     return DP
 end
 
-function signrankpdf(n::Int, W::Union{Float64,Int})
-    f, _ = modf(W)
-    if f != 0
-        return 0.0
-    end
-    W = Int(W)
-    max_W = (n * (n + 1)) >> 1
+function signrankpdf(n::Int, W::Float64)
+    return isinteger(W) ? signrankpdf(n, Int(W)) : 0.0
+end
+function signrankpdf(n::Int, W::Int)
     if W < 0
         return 0.0
-    elseif W > max_W >> 1
-        return signrankpdf(n, max_W - W)
+    end
+    max_W = (n * (n + 1)) >> 1
+    W2 = max_W - W
+    if W2 < W
+        return signrankpdf(n, W2)
     end
     DP = signrankDP(n, W)
     return ldexp(float(DP[1]), -n)
@@ -45,13 +45,17 @@ function signranklogpdf(n::Int, W::Union{Float64,Int})
     return log(signrankpdf(n, W))
 end
 
-function signrankcdf(n::Int, W::Union{Float64,Int})
-    W = round(Int, W, RoundNearestTiesUp)
-    max_W = (n * (n + 1)) >> 1
+function signrankcdf(n::Int, W::Float64)
+    return signrankcdf(n, round(Int, W, RoundNearestTiesUp))
+end
+function signrankcdf(n::Int, W::Int)
     if W < 0
         return 0.0
-    elseif W > max_W >> 1
-        return 1.0 - signrankcdf(n, max_W - W - 1)
+    end
+    max_W = (n * (n + 1)) >> 1
+    W2 = max_W - W - 1
+    if W2 < W
+        return 1.0 - signrankcdf(n, W2)
     end
     DP = signrankDP(n, W)
     return sum(Base.Fix2(ldexp, -n) ∘ float, DP)
@@ -61,10 +65,12 @@ function signranklogcdf(n::Int, W::Union{Float64,Int})
     return log(signrankcdf(n, W))
 end
 
-function signrankccdf(n::Int, W::Union{Float64,Int})
+function signrankccdf(n::Int, W::Float64)
+    return signrankccdf(n, round(Int, W, RoundNearestTiesUp))
+end
+function signrankccdf(n::Int, W::Int)
     max_W = (n * (n + 1)) >> 1
-    W = round(Int, W, RoundNearestTiesUp)
-    return signrankcdf(n, max_W - W - 1,)
+    return signrankcdf(n, max_W - W - 1)
 end
 
 function signranklogccdf(n::Int, W::Union{Float64,Int})
@@ -72,7 +78,7 @@ function signranklogccdf(n::Int, W::Union{Float64,Int})
 end
 
 function signrankinvcdf(n::Int, p::Float64)
-    if p < 0.0 || p > 1.0
+    if !(0.0 <= p <= 1.0)
         return NaN
     end
     W = 0
@@ -83,7 +89,7 @@ function signrankinvcdf(n::Int, p::Float64)
 end
 
 function signrankinvlogcdf(n::Int, logp::Float64)
-    if logp > 0.0 || logp == -Inf
+    if !(-Inf < logp <= 0.0)
         return NaN
     end
     W = 0
@@ -94,16 +100,27 @@ function signrankinvlogcdf(n::Int, logp::Float64)
 end
 
 function signrankinvccdf(n::Int, p::Float64)
-    signrankinvcdf(n, 1 - p)
+    if !(0.0 <= p <= 1)
+        return NaN
+    end
+    if p == 0.0
+        max_W = (n * (n + 1)) >> 1
+        return float(max_W)
+    end
+    W = 0
+    while signrankccdf(n, W) > p # TODO binary search and symmetry
+        W += 1
+    end
+    return float(W)
 end
 
 function signrankinvlogccdf(n::Int, logp::Float64)
-    if logp == -Inf
+    if !(-Inf < logp <= 0.0)
         return NaN
     end
-    if logp == 0
-        return 0.0
+    W = 0
+    while signranklogccdf(n, W) > logp # TODO binary search and symmetry
+        W += 1
     end
-    #signrankinvlogcdf(n, log1mexp(logp)) # does not roundtrip well
-    signrankinvccdf(n, exp(logp))
+    return float(W)
 end
--- a/test/rmath.jl
+++ b/test/rmath.jl
@@ -44,10 +44,7 @@ function rmathcomp(basename, params, X::AbstractArray, rtol=_default_rtol(params
     This slight difference causes test failures for the inverse functions,
     due to a slight shift in the location of the discontinuity.
     =#
-    test_inv = true
-    if basename == "signrank" && Sys.islinux()
-        test_inv = false
-    end
+    test_inv = basename != "signrank" || !Sys.islinux()
 
     if has_pdf
         pdf     = string(basename, "pdf")
--- a/test/rmath.jl
+++ b/test/rmath.jl
@@ -44,9 +44,9 @@ function rmathcomp(basename, params, X::AbstractArray, rtol=_default_rtol(params
     This slight difference causes test failures for the inverse functions,
     due to a slight shift in the location of the discontinuity.
     =#
-    has_inv = true
+    test_inv = true
     if basename == "signrank" && Sys.islinux()
-        has_inv = false
+        test_inv = false
     end
 
     if has_pdf
@@ -57,7 +57,7 @@ function rmathcomp(basename, params, X::AbstractArray, rtol=_default_rtol(params
     ccdf    = string(basename, "ccdf")
     logcdf  = string(basename, "logcdf")
     logccdf = string(basename, "logccdf")
-    if has_inv
+    if test_inv
         invcdf     = string(basename, "invcdf")
         invccdf    = string(basename, "invccdf")
         invlogcdf  = string(basename, "invlogcdf")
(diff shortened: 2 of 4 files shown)


andreasnoack:
I'd try to add it to test/rmath.jl` similarly to https://github.com/JuliaStats/StatsFuns.jl/blob/8f50565d0e27d3b827d744d2fc0dee8b790ed51c/test/rmath.jl#L233-L241 to compare with the Rmath results.

ArnoStrouwen:
Some differences with R remain:
How minus infinity is handled in `invlogcdf`, in my opinion is it correct to return zero here and not NaN.
```jl
julia> signrankinvcdf(4, 0.0)
0.0
julia> signrankinvlogcdf(4, log(0))
0.0
julia> qsignrank.(0.0,4,true,false) 
0.0
julia> qsignrank.(log(0.0),4,true,true)
NaN
```

Rounding in the quantiles is also an issue:
```jl
julia> qsignrank.(-2.1,4,true,true)
1.0
julia> qsignrank.(-2.0794415416798357,4,true,true)
1.0
julia> qsignrank.(-2.0,4,true,true)
2.0

julia> signrankinvlogcdf(4,-2.1)
1.0
julia> signrankinvlogcdf(4,-2.0794415416798357)
2.0
julia> signrankinvlogcdf(4,-2.0)
2.0
```
The true cdf jumps at 0.125, but that value does not roundtrip with exp/log.
```jl
julia> exp(-2.0794415416798357)
0.12500000000000003
julia> log(0.125)
-2.0794415416798357
julia> exp(log(0.125))
0.12500000000000003
```

ArnoStrouwen:
I improved rounding by getting rid of all `exp` in my code.
Still, some rounding differences with R remain for `logccdf` and `logcdf`, which I don't know how to solve.

src/distrs/signrank.jl
```diff
@@ -0,0 +1,97 @@
...
+    end
+    return float(W)
+end
+
+function signrankinvccdf(n::Int, p::Float64)
+    signrankinvcdf(n, 1 - p)
+end
+
+function signrankinvlogccdf(n::Int, logp::Float64)
+    signrankinvlogcdf(n, log1mexp(logp))
```
andreasnoack, line 96: I think that you might be able to match for this one if you instead of relying on `log1mexp` then do a search similarly to `signrankinvlogcdf`.

ArnoStrouwen:
Stuck on this: for higher `n`, the R version seems to behave weirdly?
```jl
julia> psignrank(1,4,true,false)
0.125
julia> qsignrank(0.125,4,true,false)
1.0
julia> signrankinvcdf(4,0.125)
1.0

julia> psignrank(1,50,true,false)
1.776356839400249e-15
julia> qsignrank(1.776356839400249e-15,50,true,false)
0.0
julia> signrankinvcdf(50,1.776356839400249e-15)
1.0
```

```jl
julia> qsignrank.(psignrank.(-1:8,4,true,false),4,true,false)
10-element Vector{Float64}:
 0.0
 0.0
 1.0
 2.0
 3.0
 4.0
 5.0
 6.0
 7.0
 8.0

julia> qsignrank.(psignrank.(-1:8,50,true,false),50,true,false)
10-element Vector{Float64}:
 0.0
 0.0
 0.0
 0.0
 2.0
 3.0
 5.0
 6.0
 7.0
 8.0
```

andreasnoack:
Would be good to file a bug report in the R bug tracker https://bugs.r-project.org/. Here you can avoid testing these cases against R for now and instead test some invariants such that the round tripping of p and q.

ArnoStrouwen:
Seems like windows/mac have slightly different rounding going on somewhere?

For the older julia versions, what would be the best way to round to nearest integer?

src/distrs/signrank.jl
```diff
@@ -0,0 +1,109 @@
...
+    end
+    return DP
+end
+
+function signrankpdf(n::Int, W::Union{Float64,Int})
+    f, _ = modf(W)
+    if f != 0
+        return 0.0
+    end
+    W = Int(W)
```
devmotion, line 33: A bit simpler (and possibly better supported on older Julia versions):
```suggestion
function signrankpdf(n::Int, W::Float64)
    return isinteger(W) ? signrankpdf(n, Int(W)) : 0.0
end
function signrankpdf(n::Int, W::Int)
```

src/distrs/signrank.jl
```diff
@@ -0,0 +1,109 @@
...
+    max_W = (n * (n + 1)) >> 1
+    if W < 0
+        return 0.0
+    elseif W > max_W >> 1
+        return signrankpdf(n, max_W - W)
+    end
```
devmotion, line 39: IMO a bit clearer since it's a bit more obvious that computations are always carried out for the minimum of W and max_W - W. Also slightly optimized since max_W does not have to be computed for W < 0:
```suggestion
    if W < 0
        return 0.0
    end
    max_W = (n * (n + 1)) >> 1
    W2 = max_W - W
    if W2 < W
        return signrankpdf(n, W2)
    end
```

src/distrs/signrank.jl
```diff
@@ -0,0 +1,109 @@
...
+    DP = signrankDP(n, W)
+    return ldexp(float(DP[1]), -n)
+end
+
+function signranklogpdf(n::Int, W::Union{Float64,Int})
+    return log(signrankpdf(n, W))
+end
+
+function signrankcdf(n::Int, W::Union{Float64,Int})
+    W = round(Int, W, RoundNearestTiesUp)
```
devmotion, line 49:
```suggestion
function signrankcdf(n::Int, W::Float64)
    return signrancdf(n, round(Int, W, RoundNearestTiesUp))
end
function signrankcdf(n::Int, W::Int)
```

src/distrs/signrank.jl
```diff
@@ -0,0 +1,109 @@
...
+    max_W = (n * (n + 1)) >> 1
+    if W < 0
+        return 0.0
+    elseif W > max_W >> 1
+        return 1.0 - signrankcdf(n, max_W - W - 1)
+    end
```
devmotion, line 55:
```suggestion
    if W < 0
        return 0.0
    end
    max_W = (n * (n + 1)) >> 1
    W2 = max_W - W - 1
    if W2 < W
        return 1.0 - signrankcdf(n, W2)
    end
```

src/distrs/signrank.jl
```diff
@@ -0,0 +1,109 @@
...
+    return sum(Base.Fix2(ldexp, -n) ∘ float, DP)
+end
+
+function signranklogcdf(n::Int, W::Union{Float64,Int})
+    return log(signrankcdf(n, W))
+end
+
+function signrankccdf(n::Int, W::Union{Float64,Int})
+    max_W = (n * (n + 1)) >> 1
+    W = round(Int, W, RoundNearestTiesUp)
```
devmotion, line 66: It seems a bit unnecessary to compute max_W both here and in signrankcdf.
```suggestion
function signrankccdf(n::Int, W::Float64)
    return signrankccdf(n, W, RoundNearestTiesUp)
end
function signrankccdf(n::Int, W::Int)
    max_W = (n * (n + 1)) >> 1
```

src/distrs/signrank.jl
```diff
@@ -0,0 +1,109 @@
...
+    end
+    W = 0
+    while signranklogcdf(n, W) < logp # TODO binary search and symmetry
+        W += 1
+    end
+    return float(W)
+end
+
+function signrankinvccdf(n::Int, p::Float64)
+    signrankinvcdf(n, 1 - p)
```
devmotion, line 97: Maybe use the same approach as in `signrankinvcdf` instead to avoid possible floating point issues in `1 - p`?
```suggestion
    if !(0.0 <= p <= 1)
        return NaN
    end
    W = 0
    while signrankccdf(n, W + 1) >= p # TODO binary search and symmetry
        W += 1
    end
    return float(W)
```

src/distrs/signrank.jl
```diff
@@ -0,0 +1,109 @@
...
+
+function signrankinvlogccdf(n::Int, logp::Float64)
+    if logp == -Inf
+        return NaN
+    end
+    if logp == 0
+        return 0.0
+    end
+    #signrankinvlogcdf(n, log1mexp(logp)) # does not roundtrip well
+    signrankinvccdf(n, exp(logp))
```
devmotion, line 108: Same here:
```suggestion
    if !(-Inf < logp <= 0.0)
        return NaN
    end
    W = 0
    while signranklogccdf(n, W + 1) >= logp # TODO binary search and symmetry
        W += 1
    end
    return float(W)
```

test/rmath.jl
```diff
@@ -215,7 +215,7 @@ end
             @test @inferred(betainvccdf(α, 0, p)) === 1f0
         end
     end
-
+
```
devmotion, line 218:
```suggestion

```

test/rmath.jl
```diff
@@ -376,6 +376,23 @@ end
...
+    
+    @test signrankinvcdf.(10, signrankcdf.(10, -1:56)) ≈ [0; 0:55; 55] atol = 1e-12 rtol = 1e-12
+    @test signrankinvccdf.(10, signrankccdf.(10, -1:56)) ≈ [0; 0:55; 55] atol = 1e-12 rtol = 1e-12
+    @test signrankinvlogcdf.(10, signranklogcdf.(10, -1:56)) ≈ [NaN; 0:55; 55] nans = true atol = 1e-12 rtol = 1e-12
+    @test signrankinvlogccdf.(10, signranklogccdf.(10, -1:56)) ≈ [0; 0:54; NaN; NaN] nans = true atol = 1e-12 rtol = 1e-12
+
+    @test signrankinvcdf.(50, signrankcdf.(50, -1:1276)) ≈ [0; 0:1275; 1275] atol = 1e-12 rtol = 1e-12
+    @test signrankinvccdf.(50, signrankccdf.(50, -1:1276)) ≈ [0; 0:1275; 1275] atol = 1e-12 rtol = 1e-12
+    @test signrankinvlogcdf.(50, signranklogcdf.(50, -1:1276)) ≈ [NaN; 0:1275; 1275] nans = true atol = 1e-12 rtol = 1e-12
+    @test signrankinvlogccdf.(50, signranklogccdf.(50, -1:1276)) ≈ [0; 0:1274; NaN; NaN] nans = true atol = 1e-12 rtol = 1e-12
```
devmotion, line 394: I think it would be good to compare evaluations for each value separately to avoid that differences are masked by some large values. All existing tests seem to loop over the desired inputs.

src/distrs/signrank.jl
ArnoStrouwen: Not sure about this one, I think it gets the rounding wrong then.

devmotion: I was wondering the same, but the suggestion is just copied from your current implementation - I did not change the rounding here 😄

src/distrs/signrank.jl
ArnoStrouwen: This one causes an infinite loop.

src/distrs/signrank.jl
ArnoStrouwen: This works, but placing a note to double-check due to the issues with the non-log version.

src/distrs/signrank.jl
ArnoStrouwen: It is also off by one for most cases.

ArnoStrouwen:
This failure is unique to ubuntu:
https://github.com/JuliaStats/StatsFuns.jl/actions/runs/13596765039/job/38015305607?pr=173#step:6:1507
Not present on my windows machine or the windows testing machine.
Can somebody with linux verify?

src/distrs/signrank.jl
```diff
@@ -0,0 +1,126 @@
...
+function signranklogccdf(n::Int, W::Union{Float64,Int})
+    return log(signrankccdf(n, W))
+end
+
+function signrankinvcdf(n::Int, p::Float64)
+    if !(0.0 <= p <= 1.0)
+        return NaN
+    end
+    W = 0
+    while signrankcdf(n, W) < p # TODO binary search and symmetry
```
devmotion, line 85: Can you open an issue for the remaining TODOs, to ensure they're not forgotten once the PR is merged?

test/rmath.jl
```diff
@@ -34,6 +34,20 @@ function rmathcomp(basename, params, X::AbstractArray, rtol=_default_rtol(params
...
+    -0.2076393647782445
+    julia> psignrank(18,10,false,true) # linux
+    -0.20763936477824452
+    This slight difference causes test failures for the inverse functions,
+    due to a slight shift in the location of the discontinuity.
+    =#
+    has_inv = true
+    if basename == "signrank" && Sys.islinux()
+        has_inv = false
+    end
```
devmotion, line 50: Maybe `test_inv` would be more appropriate since the inv functions exist in all cases.
```suggestion
    test_inv = basename != "signrank" || !Sys.islinux()
```

test/rmath.jl
```diff
@@ -43,10 +57,12 @@ function rmathcomp(basename, params, X::AbstractArray, rtol=_default_rtol(params
     ccdf    = string(basename, "ccdf")
     logcdf  = string(basename, "logcdf")
     logccdf = string(basename, "logccdf")
-    invcdf     = string(basename, "invcdf")
-    invccdf    = string(basename, "invccdf")
-    invlogcdf  = string(basename, "invlogcdf")
-    invlogccdf = string(basename, "invlogccdf")
+    if has_inv
```
devmotion, line 60:
```suggestion
    if test_inv
```

test/rmath.jl
```diff
@@ -376,6 +398,29 @@ end
         ((Inf32,), -5f0:0.1f0:5f0),
     ])
 
+    rmathcomp_tests("signrank", [
+        ((4),-2:12),
+        ((4),-2.0:0.25:12.0),
+        ((10),-2:57),
+        #((50),-2:1277),
```
devmotion, line 405: Should this be removed? Or uncommented?

test/rmath.jl
```diff
@@ -70,10 +88,12 @@ function rmathcomp(basename, params, X::AbstractArray, rtol=_default_rtol(params
     rmath_ccdf    = get_rmathfun(ccdf)
     rmath_logcdf  = get_rmathfun(logcdf)
     rmath_logccdf = get_rmathfun(logccdf)
-    rmath_invcdf     = get_rmathfun(invcdf)
-    rmath_invccdf    = get_rmathfun(invccdf)
-    rmath_invlogcdf  = get_rmathfun(invlogcdf)
-    rmath_invlogccdf = get_rmathfun(invlogccdf)
+    if has_inv
```
devmotion, line 91:
```suggestion
    if test_inv
```

test/rmath.jl
```diff
@@ -57,10 +73,12 @@ function rmathcomp(basename, params, X::AbstractArray, rtol=_default_rtol(params
     stats_ccdf    = get_statsfun(ccdf)
     stats_logcdf  = get_statsfun(logcdf)
     stats_logccdf = get_statsfun(logccdf)
-    stats_invcdf     = get_statsfun(invcdf)
-    stats_invccdf    = get_statsfun(invccdf)
-    stats_invlogcdf  = get_statsfun(invlogcdf)
-    stats_invlogccdf = get_statsfun(invlogccdf)
+    if has_inv
```
devmotion, line 76:
```suggestion
    if test_inv
```

test/rmath.jl
```diff
@@ -114,21 +134,23 @@ function rmathcomp(basename, params, X::AbstractArray, rtol=_default_rtol(params
...
-            params, "q", _p, false, rtol)
-    end
-    @testset "invlogcdf with log(q)=$_p" for _p in lp
-        check_rmath(invlogcdf, stats_invlogcdf, rmath_invlogcdf,
-            params, "lq", _p, false, rtol)
-    end
-    @testset "invlogccdf with log(q)=$_p" for _p in lcp
-        check_rmath(invlogccdf, stats_invlogccdf, rmath_invlogccdf,
-            params, "lq", _p, false, rtol)
+    if has_inv
```
devmotion, line 137:
```suggestion
    if test_inv
```

test/rmath.jl
```diff
@@ -376,6 +398,29 @@ end
...
+    @test isnan(signrankinvlogccdf.(10, signranklogccdf.(10, 55)))
+    @test isnan(signrankinvlogccdf.(10, signranklogccdf.(10, 56)))
+
+    @test signrankinvcdf.(50, signrankcdf.(50, -1:1276)) == [0; 0:1275; 1275]
+    @test signrankinvccdf.(50, signrankccdf.(50, -1:1276)) == [0; 0:1275; 1275]
+    @test signrankinvlogcdf.(50, signranklogcdf.(50, 0:1276)) == [0:1275; 1275]
+    @test isnan(signrankinvlogcdf.(50, signranklogcdf(500, -1)))
+    @test signrankinvlogccdf.(50, signranklogccdf.(50, -1:1274)) == [0; 0:1274]
+    @test isnan(signrankinvlogccdf.(50, signranklogccdf.(50, 1275)))
+    @test isnan(signrankinvlogccdf.(50, signranklogccdf.(50, 1276)))
```
devmotion, line 422: Are these tests not covered by the default test suite above?

devmotion (commented):
I came across https://aakinshin.net/posts/signrank-limitations/ a while ago - are we doing better here or do we face the same/similar problems?

test/rmath.jl
ArnoStrouwen: Maybe I'll switch it to a broken test, so we can easily track if R fixes it behavior.

ArnoStrouwen:
> I came across https://aakinshin.net/posts/signrank-limitations/ a while ago - are we doing better here or do we face the same/similar problems?

We are definitely doing better in some cases, hence the inability to test vs R in all cases.
There is a limit to how well you can do for large n since the probabilities in the tails become incredibly small.

The pdf, for example, is as accurate as `float` and `ldexp` allow.
`ldexp(float(DP[1]), -n)`
The result in DP is the exact integer value.
The only other source of error is that DP might overflow, but I don't think that is the issue in the link you posted.

Not all functions are as accurate as they could be, there are a few inconsistencies carried over from the R versions:
signrank uses `log(1-p)`, while wilcox uses the more accurate `log1p`.

ArnoStrouwen:
If there are no further comments, I would like this to get merged into to unblock upstream work.

andreasnoack:
@devmotion do you have further comments here?

test/rmath.jl
```diff
@@ -34,6 +34,20 @@ function rmathcomp(basename, params, X::AbstractArray, rtol=_default_rtol(params
...
+    -0.2076393647782445
+    julia> psignrank(18,10,false,true) # linux
+    -0.20763936477824452
+    This slight difference causes test failures for the inverse functions,
+    due to a slight shift in the location of the discontinuity.
+    =#
+    test_inv = true
+    if basename == "signrank" && Sys.islinux()
+        test_inv = false
+    end
```
devmotion, line 50: Maybe simplify this to
```suggestion
    test_inv = basename != "signrank" || !Sys.islinux()
```
?

devmotion (approved):
Looks good to me.

devmotion (approved):

PR #174, opened 2025-02-23
add wilcox distribution

ArnoStrouwen:
Adds `wilcox` distribution.
~~Still a work in progress.~~


ArnoStrouwen:
This is what is causing test failures:
```jl
julia> psignrank(18,10,false,true) # windows
-0.2076393647782445
julia> psignrank(18,10,false,true) # linux
-0.20763936477824452
```

ArnoStrouwen:
This one is also ready for review @devmotion.

src/StatsFuns.jl
```diff
@@ -228,6 +228,18 @@ export
     srdistinvccdf,       # inverse-ccdf of studentized range distribution
     srdistinvlogcdf,     # inverse-logcdf of studentized range distribution
     srdistinvlogccdf,    # inverse-logccdf of studentized range distribution
+    
+    # distrs/wilcox
```
devmotion, line 232: Can you add these to the README as well? (I notice now that we forgot to do that in #173, maybe fix it here as well or open a separate PR for `signrank`?)

src/distrs/wilcox.jl
```diff
@@ -0,0 +1,155 @@
+#=
+The wilcox distribution is equivalent to the problem
+of finding the number of subsets with size nx
+of {1,2,...,nx,nx+1,...,nx+ny} summing to (U-sum(1:nx))
+relative to the total number of subsets with size nx.
+The empty subset is defined to sum to zero.
+This can be calculated using the recursion:
+either nx+ny is in the subset in witch case we need to calculate
```
devmotion, line 8:
```suggestion
either nx+ny is in the subset in which case we need to calculate
```

src/distrs/wilcox.jl
```diff
@@ -0,0 +1,155 @@
...
+the number of subsets with size k of {1,2,...,k,k+1,...,k+j-1} summing to i.
+ =#
+
+@inline function wilcoxDP(nx, ny, U)
+    DP = zeros(Int, sum(1:nx)+U, ny + 1)
+    for i in 1:(nx+ny)
+        for j in i:(ny+1)
+            DP[i, j] = 1
+        end
+    end
```
devmotion, line 26: Maybe swap the iteration order to make it a bit friendlier to Julia's column-major ordering?
```suggestion
    for j in 1:(ny+1)
        for i in 1:j
            DP[i, j] = 1
        end
    end
```
(This would also automatically skip i > ny + 1, in which case currently the inner loop is empty)

src/distrs/wilcox.jl
```diff
@@ -0,0 +1,155 @@
...
+    for k in 2:nx
+        for j in 1:(ny+1)
+            for i in min(U + sum(1:k), sum((j):(k+j-1))):-1:1
```
devmotion, line 29: I think it should be possible to minimize calculations here a bit:
- We know that `sum(1:k) = (k * (k + 1)) >> 1`, and it could be calculated only once in the outer loop
- Moreover, we know that `sum(j:(k+j-1)) = sum(1:(k+j-1)) - sum(1:(j-1)) = (k * (k + 2 * j - 1)) >> 1`

We even know that `min(U + sum(1:k), sum(j:(k+j-1))) = min(U + sum(1:k), sum(1:k) + k * (j - 1)) = sum(1:k) + min(U, k * (j - 1))`.
Thus we could
a) Precompute `a = (k * (k + 1)) >> 1` in the outer loop (possibly `a` is not a good variable name)
b) Increment `b = b + k` with initial value `b = 0` after every iteration of `for j ...` (could be skipped as soon as `b >= U`)
c) Use
```suggestion
            for i in (a + min(U, b)):-1:1
```
here

src/distrs/wilcox.jl
```diff
@@ -0,0 +1,155 @@
...
+                if j > 1
```
devmotion, line 30: Maybe the iteration over `j` could be split in two parts, `j = 1` and `for j in 2:(ny+1)`, to avoid this check?

src/distrs/wilcox.jl
```diff
@@ -0,0 +1,155 @@
...
+                    DP[i, j] = DP[i, j-1]
+                else
+                    DP[i, j] = 0
+                end
+                if i - j - k + 1 >= 1
```
devmotion, line 35: This is just
```suggestion
                if i >= j + k
```
As above, maybe just split iteration over `i` in two parts, `for i in (a + min(U, b)):-1:(j + k)` and `for i in (j + k - 1):-1:1`?

src/distrs/wilcox.jl
```diff
@@ -0,0 +1,155 @@
...
+end
+
+function wilcoxpdf(nx::Int, ny::Int, U::Float64)
+    return isinteger(U) ? wilcoxpdf(nx, ny, Int(W)) : 0.0
+end
+function wilcoxpdf(nx::Int, ny::Int, U::Int)
+    if U < 0
+        return 0.0
+    end
+    max_U = sum((ny+1):(ny+nx)) - sum(1:nx)
```
devmotion, line 51: Probably should use the explicit formula here as well:
```suggestion
    max_U = nx * ny
```

src/distrs/wilcox.jl
```diff
@@ -0,0 +1,155 @@
...
+    U2 = max_U - U
+    if U2 < U
+        return wilcoxpdf(nx, ny, U2)
+    end
+    DP = wilcoxDP(nx, ny, U)
+    DP[sum(1:nx)+U, ny+1] / binomial(nx+ny, nx)
+end
+
+function wilcoxlogpdf(nx::Int, ny::Int, U::Union{Float64,Int})
+    return log(wilcoxpdf(nx, ny, U))
```
devmotion, line 61: Maybe it would be numerically advantageous to use `log(DP[...]) - logbinomial(...)` instead of `log(DP[...]/binomial(...))` in certain cases?

src/distrs/wilcox.jl
```diff
@@ -0,0 +1,155 @@
...
+    if U < 0
+        return 0.0
+    end
+    max_U = sum((ny+1):(ny+nx)) - sum(1:nx)
+    U2 = max_U - U
+    if U2 < U
+        return wilcoxpdf(nx, ny, U2)
+    end
+    DP = wilcoxDP(nx, ny, U)
+    DP[sum(1:nx)+U, ny+1] / binomial(nx+ny, nx)
```
devmotion, line 57:
```suggestion
    DP[(((nx * (nx + 1)) >> 1)+U, ny+1] / binomial(nx+ny, nx)
```

src/distrs/wilcox.jl
```diff
@@ -0,0 +1,155 @@
...
+end
+
+function wilcoxcdf(nx::Int, ny::Int, U::Float64)
+    wilcoxcdf(nx, ny, round(Int, U, RoundNearestTiesUp))
+end
+function wilcoxcdf(nx::Int, ny::Int, U::Int)
+    if U < 0
+        return 0.0
+    end
+    max_U = sum((ny+1):(ny+nx)) - sum(1:nx)
```
devmotion, line 71:
```suggestion
    max_U = nx * ny
```

src/distrs/wilcox.jl
```diff
@@ -0,0 +1,155 @@
...
+    U2 = max_U - U - 1
+    if U2 < U
+        return 1.0 - wilcoxcdf(nx, ny, U2)
+    end
+    DP = wilcoxDP(nx, ny, U)
+    sum(DP[(sum(1:nx)):(sum(1:nx)+U), ny+1] ./ binomial(nx+ny, nx))
```
devmotion, line 77:
```suggestion
    z = (nx * (nx + 1)) >> 1
    sum(@view(DP[z:(z+U), ny+1])) / binomial(nx+ny, nx)
```

src/distrs/wilcox.jl
```diff
@@ -0,0 +1,155 @@
...
+end
+
+function wilcoxlogcdf(nx::Int, ny::Int, U::Union{Float64,Int})
+    max_U = sum((ny+1):(ny+nx)) - sum(1:nx)
```
devmotion, line 81:
```suggestion
    max_U = nx * ny
```

src/distrs/wilcox.jl
```diff
@@ -0,0 +1,155 @@
...
+    else
+        return log(wilcoxcdf(nx, ny, U))
+    end
+end
+
+function wilcoxccdf(nx::Int, ny::Int, U::Float64)
+    wilcoxccdf(nx, ny, round(Int, U, RoundNearestTiesUp))
+end
+function wilcoxccdf(nx::Int, ny::Int, U::Int)
+    max_U = sum((ny+1):(nx+ny)) - sum(1:nx)
```
devmotion, line 94:
```suggestion
    max_U = nx * ny
```

src/distrs/wilcox.jl
```diff
@@ -0,0 +1,155 @@
...
+    wilcoxcdf(nx, ny, max_U - U - 1)
+end
+
+function wilcoxlogccdf(nx::Int, ny::Int, U::Union{Float64,Int})
+    max_U = sum((ny+1):(ny+nx)) - sum(1:nx)
```
devmotion, line 99:
```suggestion
    max_U = nx * ny
```

src/distrs/wilcox.jl
```diff
@@ -0,0 +1,155 @@
...
+    U2 = max_U - U - 1
+    if U2 < U+2 # +2 is emperical
```
devmotion, line 101:
```suggestion
    if U2 < U+2 # +2 is empirical
```

src/distrs/wilcox.jl
```diff
@@ -0,0 +1,155 @@
...
+    end
+    return float(U)
+end
+
+function wilcoxinvccdf(nx::Int, ny::Int, p::Float64)
+    if !(0.0 <= p <= 1)
+        return NaN
+    end
+    if p == 0.0
+        max_U = sum((ny+1):(ny+nx)) - sum(1:nx)
```
devmotion, line 136:
```suggestion
        max_U = nx * ny
```

src/distrs/wilcox.jl
ArnoStrouwen: Unrelated, but do you find the comment clear about what is being calculated?

src/distrs/wilcox.jl
ArnoStrouwen: I agree that it is not the most elegant code,
but splitting the loop would require you to also duplicate the lines:
```jl
                if i - j - k + 1 >= 1
                    DP[i, j] += DP[i-j-k+1, j]
                end
```
which I'm also not keen on.

ArnoStrouwen: What would be ideal is if instead of increasing number of zero rows at the top of DP as k increases, could be "reclaimed" to do useful work, then the first column would alwyas just have a one as the first element and zero elsewhere and we could just start iterating at the second column. However, I can't make the math work out.
k=1
```
 1  1  1  1  1
 0  1  1  1  1
 0  0  1  1  1
 0  0  0  1  1
 0  0  0  0  1
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0
```
k=2, notice the two useless zero rows at the top.
```
 0  0  0  0  0
 0  0  0  0  0
 1  1  1  1  1
 0  1  1  1  1
 0  1  2  2  2
 0  0  1  2  2
 0  0  1  2  3
 0  0  0  1  2
 0  0  0  1  2
 0  0  0  0  1
 0  0  0  0  1
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0
```

src/distrs/wilcox.jl
ArnoStrouwen: I took a middle ground between performance and readability.

src/distrs/wilcox.jl
ArnoStrouwen: I'm unsure, I followed the same rounding techniques that R uses.

ArnoStrouwen:
@devmotion, I think I succeeded in expressing the loops a bit more elegantly. Please give this another round of review.

src/distrs/wilcox.jl
```diff
@@ -0,0 +1,149 @@
...
+        DP[1, j] = 1
+    end
+    for k = 1:nx
+        for j = 2:(ny+1)
+            for i = (U+1):-1:2
+                DP[i, j] = DP[i, j-1]
+                if i - j >= 0
+                    DP[i, j] += DP[i-j+1, j]
+                end
+            end
```
devmotion, line 32: I'd use the version without `if`, also to reduce memory accesses to `DP[i, j]`:
```suggestion
            for i = (U+1):-1:j
                DP[i, j] = DP[i-j+1, j] + DP[i, j-1] 
            end
            for i = (j-1):-1:2
                DP[i, j] = DP[i, j-1] 
            end
```

src/distrs/wilcox.jl
```diff
@@ -0,0 +1,149 @@
...
+    if U < 0
+        return 0.0
+    end
+    max_U = nx * ny
+    U2 = max_U - U - 1
+    if U2 < U
+        return 1.0 - wilcoxcdf(nx, ny, U2)
+    end
+    DP = wilcoxDP(nx, ny, U)
+    sum(@view(DP[1:(U+1), ny+1]) ./ binomial(nx + ny, nx))
```
devmotion, line 71: To avoid an intermediate array:
```suggestion
    sum(@view(DP[1:(U+1), ny+1])) / binomial(nx + ny, nx)
```

test/rmath.jl
```diff
@@ -386,7 +386,21 @@ end
...
+    ])
+    
+    
+    @test wilcoxinvcdf.(10, 10, wilcoxcdf.(10, 10, -1:101)) == [0; 0:100; 100]
+    @test wilcoxinvccdf.(10, 10, wilcoxccdf.(10, 10, -1:101)) == [0; 0:100; 100]
+    @test wilcoxinvlogcdf.(10, 10, wilcoxlogcdf.(10, 10, 0:101)) == [0:100; 100]
+    @test isnan(wilcoxinvlogcdf.(10, 10, wilcoxlogcdf(10, 10, -1)))
+    @test wilcoxinvlogccdf.(10, 10, wilcoxlogccdf.(10, 10, -1:99)) == [0; 0:99]
+    @test isnan(wilcoxinvlogccdf.(10, 10, wilcoxlogccdf.(10, 10, 100)))
+    @test isnan(wilcoxinvlogccdf.(10, 10, wilcoxlogccdf.(10, 10, 101)))
```
devmotion, line 402: Maybe add more comparisons to Rmath? And also test `Float64` inputs for `wilcoxcdf`, `wilcoxpdf` etc.?

src/distrs/wilcox.jl
ArnoStrouwen: `(U+1)` is a loose upper bound for the iteration of `i`, do you know the correct tight bound?
I think it is something like:
`for i = min((U+1),sum((j):(k+j-1))-sum(1:k)+1):-1:j`

src/distrs/wilcox.jl
ArnoStrouwen: Included it, but doesn't seem to make a difference to allocations on my machine.

test/rmath.jl
ArnoStrouwen: Good shout, this caught a bug.

src/distrs/wilcox.jl
ArnoStrouwen: The splitting of the loop doesn't work, because the start of the second loop can be before the start of the first, i.e. it is not guaranteed that `j` lies between `(U+1)` and `2`.
e.g. if DP is a 1 by 5 matrix, then the second loop will go out of bounds.

devmotion: > do you know the correct tight bound?

I haven't thought more about it yet 😄 

> The splitting of the loop doesn't work, because the start of the second loop can be before the start of the first, i.e. it is not guaranteed that j lies between (U+1) and 2.

Ah, my bad, I think it should have been
```suggestion
            for i = (U+1):-1:max(j,2)
                # In this loop: U+1 >= i >= 2 AND i - j >= 0
                DP[i, j] = DP[i-j+1, j] + DP[i, j-1] 
            end
            for i = min(U+1,j-1):-1:2
                # In this loop: U+1 >= i >= 2 AND i - j < 0
                DP[i, j] = DP[i, j-1] 
            end
```

devmotion: > do you know the correct tight bound?

As you wrote in the comment in the code,

> The (i,j)'th element of DP in the k'th outer loop iteration represents:
the number of subsets with size k of {1,2,...,k,k+1,...,k+j-1} summing to i+sum(1:k)-1.

Let q be the maximum sum of a subset of size k of {1,2,...,k+j-1} that may be attained. We know about q:
1. q is less than or equal to the sum of the subset {j,j+1,...,k+j-1} (k largest values in the set), ie. q <= k*(2*j+k-1)/2.
2. Any subset of size nx that consists of the subset that sums up to q and a subset of size nx-k of {k+j,k+j+1,...,nx+ny} sums up to at least q+sum((k+j):(nx+j-1)) (nx-k smallest numbers in {k+j,k+j+1,...}). Since we only consider sets of size nx that sum up to U+sum(1:nx), this implies that q+sum((k+j):(nx+j-1)) <= U+sum(1:nx), or equivalently q <= U+k*(2*j+k-1)/2-nx*j.

Taken, together we have q <= min(k*(2*j+k-1)/2, U+k*(2*j+k-1)/2-nx*j) = k*(2*j+k-1)/2+min(U-nx*j,0).

Thus i+sum(1:k)-1 <= q <= k*(2*j+k-1)/2+min(U-nx*j,0), and hence i <= (j-1)*k+1+min(U-nx*j,0).

ArnoStrouwen: Ok, I think I included all this in the latest commit.
Please do a final review of the tricky indexing here, and then hopefully this is good to go.

src/distrs/wilcox.jl
```diff
@@ -0,0 +1,152 @@
...
+    end
+    for k = 1:nx
+        for j = 2:(ny+1)
+            local_max = k * (j - 1)
+            for i = min(U + 1, local_max + 1):-1:max(j, 2)
+                # In this loop: min(U+1,local_max+1) >= i >= 2 AND i - j >= 0
+                DP[i, j] = DP[i-j+1, j] + DP[i, j-1]
+            end
+            for i = min(U + 1, local_max + 1, j - 1):-1:2
+                # In this loop: min(U+1,local_max+1) >= i >= 2 AND i - j < 0
```
devmotion, line 33: Given https://github.com/JuliaStats/StatsFuns.jl/pull/174#discussion_r2019610941, shouldn't we use
```suggestion
            i_max = k * (j - 1) + 1 + min(U - nx * j, 0)
            for i = i_max:-1:max(j, 2)
                # In this loop: i_max >= i >= 2 AND i - j >= 0
                DP[i, j] = DP[i-j+1, j] + DP[i, j-1]
            end
            for i = min(i_max, j - 1):-1:2
                # In this loop: i_max >= i >= 2 AND i - j < 0
```

ArnoStrouwen: This doesn't quite get the indexing right, rolling back to the previous version
https://github.com/JuliaStats/StatsFuns.jl/actions/runs/14151899528

devmotion: I think I made an error in my reasoning, re-reading it I think condition 2 doesn't make sense: It mixes maximum sum of a set of size k with the minimum sum of a set of size nx-k.

Maybe there's some additional improvement possible, but to derive a bound here it's certainly safe to only use information about whether there's a subset of size `k` of `{1,...,k+j-1}` that sums to `i+sum(1:k)-1`, and disregard whether such a subset can be used as part of a subset of size nx that sums up to `U + nx*(nx+1)/2` (where in the cdf calculations we're even interested in all sums `nx*(nx+1)/2, 1+nx*(nx+1)/2, ..., U+nx*(nx+1)/2`).

Generally, we know that the sum `s` of a subset of size `k` of `{1,...,k+j-1}` satisfies

$$
\frac{k(k+1)}{2} \leq s \leq \frac{k(2j+k-1}{2}
$$

(sum of the `k` smallest and `k` largest values).
Thus we know that the number of subsets that sum to `i+sum(1:k)-1 = i+k(k+1)/2-1` is zero if

$$
i + \frac{k(k+1)}{2} - 1 < \frac{k(k+1)}{2} \lor i + \frac{k(k+1)}{2} - 1 > \frac{k(2j+k-1}{2},
$$

one if

$$
i + \frac{k(k+1)}{2} - 1 = \frac{k(k+1)}{2} \lor i + \frac{k(k+1)}{2} - 1 = \frac{k(2j+k-1}{2},
$$

and non-zero for other values of `i`.

Thus, in the `k`th outer loop
1. `DP[i,j] = 0` if `i < 1` or `i > k*(j-1) + 1`,
2. `DP[i,j] = 1` if `i = 1` or `i = k*(j-1) + 1`,
3. `DP[i,j] > 0` otherwise.

Thus we can restrict the inner loop to `2:(k*(j-1)+1)` (`i = 1` is handled in the separate loop above). Which is already done in the latest commit in this PR.

To reduce computing the max. value for `i` multiple times, I'd simplify it to

```suggestion
            i_max = min(U, k * (j - 1)) + 1
            for i = i_max:-1:max(j, 2)
                # In this loop: i_max >= i >= 2 AND i - j >= 0
                DP[i, j] = DP[i-j+1, j] + DP[i, j-1]
            end
            for i = min(i_max, j - 1):-1:2
                # In this loop: i_max >= i >= 2 AND i - j < 0
```

src/distrs/wilcox.jl
```diff
@@ -0,0 +1,152 @@
...
+    end
+    return
+end
+
+function wilcoxinvcdf(nx::Int, ny::Int, p::Float64)
+    if !(0.0 <= p <= 1.0)
+        return NaN
+    end
+    U = 0
+    while wilcoxcdf(nx, ny, U) < p # TODO binary search and symmetry
```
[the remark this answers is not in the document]
ArnoStrouwen: Yes, R does this via a global variable.

However, the `inverse` functions of signrank and wilcox seem to have very little practical application.
If they did, the R version likely would not be so buggy.
For this reason, I do not want to implement this in this PR.

I am happy with only an optimized `cdf` function, which is what is needed upstream.

I've added this comment to the issue:
https://github.com/JuliaStats/StatsFuns.jl/issues/175

src/distrs/wilcox.jl
```diff
@@ -0,0 +1,152 @@
...
+    if U < 0
+        return 0.0
+    end
+    max_U = nx * ny
+    U2 = max_U - U - 1
+    if U2 < U
+        return 1.0 - wilcoxcdf(nx, ny, U2)
+    end
+    DP = wilcoxDP(nx, ny, U)
+    sum(@view(DP[1:(U+1), ny+1]) / binomial(nx + ny, nx))
```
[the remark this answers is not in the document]
devmotion: My main problem with the current version in the PR is that it creates unnecessary allocations. It seems unnecessary to make a new array here. If you want to apply the division before summing you could also do something like
```suggestion
    sum(Base.Fix2(/, binomial(nx + ny, nx)), @view(DP[1:(U+1), ny+1]))
```
but in that case you'll still pay the price of `U+1` divisions instead of a single one, which is why I assumed `sum(float, ...)` might be the most performant approach.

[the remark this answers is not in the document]
ArnoStrouwen: I tried for `nx=30,ny=29,U=nx*ny-10`, and it gave the same floating point number.

ArnoStrouwen:
If there are no further comments, let's merge this in and release a new version such that upstream gets unblocked.

The remaining CI failure is unrelated to the PR.

devmotion (commented):
Did you benchmark it against Rmath? It would be good to avoid (major) performance regressions.

ArnoStrouwen:
```jl
julia> @btime pwilcox.(-2:13,3,4) ;
  2.500 μs (2 allocations: 192 bytes)

julia> @btime wilcoxcdf.(3,4,-2:13) ;
  1.170 μs (26 allocations: 2.72 KiB)
```

test/rmath.jl
```diff
@@ -43,8 +43,10 @@ function rmathcomp(basename, params, X::AbstractArray, rtol=_default_rtol(params
     -0.20763936477824452
     This slight difference causes test failures for the inverse functions,
     due to a slight shift in the location of the discontinuity.
+    
+    This also holds true for wilcox.
     =#
-    test_inv = basename != "signrank" || !Sys.islinux()
+    test_inv = basename != "signrank" && basename != "wilcox" || !Sys.islinux()
```
devmotion, line 49:
```suggestion
    test_inv = (basename != "signrank" && basename != "wilcox") || !Sys.islinux()
```

devmotion (approved):

PR #179, opened 2025-04-01
Add downstream tests of HypothesisTests

devmotion:
Now that signrank and wilcox distributions are about to be released, it will become possible to replace the Rmath versions in HypothesisTests with these Julia functions. Hence it might be interesting/more relevant to also run downstream tests with HypothesisTests.

Additionally, due to a discussion with @andreasnoack I started to wonder whether it would be better to run downstream tests with the latest released versions of the downstream packages instead of the development branch. One could argue that downstream packages are responsible for testing their development versions appropriately, and its our main responsibility in StatsFuns to not break user code, ie not break compatibility with existing releases.

### DIFF
--- a/.github/workflows/IntegrationTest.yml
+++ b/.github/workflows/IntegrationTest.yml
@@ -43,7 +43,7 @@ jobs:
             Pkg.develop(PackageSpec(path="."))  # resolver may fail with main deps
             Pkg.add("${{ matrix.package }}")
             Pkg.update()
-            Pkg.test()  # resolver may fail with test time deps
+            Pkg.test("${{ matrix.package }}") # resolver may fail with test time deps
           catch err
             err isa Pkg.Resolve.ResolverError || rethrow()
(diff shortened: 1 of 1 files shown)


andreasnoack (approved):

andreasnoack:
> I started to wonder whether it would be better to run downstream tests with the latest released versions of the downstream packages instead of the development branch.

Just to be clear: we are currently testing with the development version and that doesn't change with the PR?

devmotion:
> Just to be clear: we are currently testing with the development version and that doesn't change with the PR?

We are currently testing with the development version (we check out the downstream repo) but in its current form the PR changes it to installing the downstream package (`Pkg.add(...)`) instead.

andreasnoack:
Ah. Now I see it. Is this one ready? (Still marked as draft)

devmotion:
Yes, it's ready.

PR #180, opened 2025-04-01
Require Julia >= 1.10

devmotion:
I think it's safe to remove support for Julia < 1.10 by now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
           - windows-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -34,14 +34,11 @@ jobs:
           - ubuntu-latest
           - macos-latest
           - windows-latest
-        arch:
-          - x64
```
devmotion, line 38: Currently, CI on macos-latest does not use the Apple silicon but the x64 binaries. This change fixes it, now CI uses the native binaries on each runner.

andreasnoack (approved):

PR #181, opened 2025-04-18
Add JuliaFormatter

devmotion:
Same as https://github.com/JuliaStats/GLM.jl/pull/603. I used YAS style to ensure a consistent style throughout JuliaStats.

Edit: I changed to the default style, see https://github.com/JuliaStats/StatsFuns.jl/pull/181#discussion_r2051653407.


ext/StatsFunsChainRulesCoreExt.jl
```diff
@@ -2,85 +2,55 @@ module StatsFunsChainRulesCoreExt
...
-        - x * b,
-    ),
-)
+using ChainRulesCore: ChainRulesCore
+
+ChainRulesCore.@scalar_rule(betalogpdf(α::Real, β::Real, x::Number),
+                            @setup(z = digamma(α + β)),
+                            (log(x) + z - digamma(α),
+                             log1p(-x) + z - digamma(β),
+                             (α - 1) / x + (1 - β) / (1 - x)),)
```
andreasnoack, line 11: This is just so much worse. I'll propose that we just use JuliaFormatter's default formatting. Since the formatting is handled automatically, it shouldn't matter much that it is different between packages.

devmotion: Indeed, these lines are particularly bad IMO. I changed to the default style. As an alternative and additional comparison, I opened https://github.com/JuliaStats/StatsFuns.jl/pull/182 which uses Runic instead of JuliaFormatter for formatting the code.

devmotion:
Superseded by #182.

PR #209, opened 2026-04-05
Convert tests to TestItems framework

andreasnoack:
## Summary
- Migrate all tests from `@testset`/`include()` to `@testitem` blocks (TestItems.jl) with `TestItemRunner.jl` for execution
- Shared test utilities (`_default_rtol`) wrapped in a `@testmodule Utils` setup block
- Each test file is now a self-contained `@testitem` with its own `using` statements

## Test plan
- [x] `Pkg.test()` passes (1,166,082 tests pass, 2 expected broken)


test/inverse.jl
```diff
@@ -1,11 +1,13 @@
-using StatsFuns
-using Test
-using InverseFunctions
+@testitem "Inverse" tags = [:autodiff] begin
```
devmotion: These are not AD tests, are they?

```suggestion
@testitem "Inverse" begin
```

test/utils.jl
```diff
@@ -1,14 +1,18 @@
-# default relative tolerance for comparisons with Rmath
-function _default_rtol(params, X::AbstractArray)
-    # has to take into account `params` as well since otherwise e.g. `X::Array{<:Rational}`
-    # always uses a tolerance based on `eps(one(Float64))` even when parameters are of type
-    # Float32
-    return _default_rtol(float(promote_type(Base.promote_typeof(params...), eltype(X))))
-end
+@testmodule Utils begin
```
devmotion, line 1: Why is this a `@testmodule` but the Rmath helpers above are `@testsnippets`? Also, aren't the `RmathHelpers` copying some of the code here?

test/inverse.jl
andreasnoack: Good catch — these are `InverseFunctions` round-trips, not AD. Dropped the tag in 837cedb.

_— drafted by Claude Code (on Andreas's behalf)_

test/utils.jl
andreasnoack: Both fixed in d81eb53. `RMathHelpers` was a `@testsnippet` only because `rmathcomp` looked functions up via `getproperty(@__MODULE__, …)`, which needs the code spliced into each testitem. Those functions all live in `StatsFuns`, so I switched to `getproperty(StatsFuns, …)` and moved the helpers into the `Utils` `@testmodule` — so it's now one module with a single `_default_rtol`, no snippet.

_— drafted by Claude Code (on Andreas's behalf)_

devmotion (approved):

PR #213, opened 2026-05-29
Replace CompatHelper with Dependabot for Julia packages

devmotion:


andreasnoack:
Is this the modern approach

DilumAluthge:
CompatHelper is still maintained (for bug fixes and security issues), but we have begun encouraging folks to migrate to Dependabot, because:
1. It lets us offload a lot of the work onto GitHub.
2. It lets folks use the same tool (Dependabot) for multiple language ecosystems.

devmotion:
The CompatHelper repo tells users

> CompatHelper is now in maintenance mode. We recommend that users migrate to GitHub Dependabot

dependabot doesn't support private registries but for public packages it seems sufficient, eg DifferentiationInterface switched a while ago.

DilumAluthge:
> dependabot doesn't support private registries

True, but I'm hoping that we'll be able to add support for that in the future.

andreasnoack (approved):

PR #214, opened 2026-06-09
add owens_t function

### LINKED ISSUE #99: Feature request: Owen's t function
[Owen's T function](https://en.wikipedia.org/wiki/Owen%27s_T_function) currently does not seem to have a native julia implementation. Most existing implementations in Fortran or C are under GPL, which makes this a little tricky.

This would be useful for the [Skew Normal distribution in Distributions.jl](https://github.com/JuliaStats/Distributions.jl/pull/1104).

stevengj:
Closes #99, with a port of the well-tested and fairly standard `Float64` implementation from Boost.  *Update:* including a chain rule for the derivatives.

(Various other algorithms were proposed in that issue, but I think it makes sense to start with a trusted baseline.)


stevengj:
Unfortunately, the Boost tests don't seem to cover all of the branches, so I'll need to add more.

src/owens_t.jl
devmotion, line 1: Can you move this to `src/owens_t.jl`? `src/distrs/` contains (log)pdf, (log)cdf etc. functions of specific distributions whereas this file seems similar to https://github.com/JuliaStats/StatsFuns.jl/blob/master/src/tvpack.jl

src/distrs/owens_t.jl
```diff
@@ -0,0 +1,247 @@
...
+#      Use, modification and distribution are subject to the
+#      Boost Software License, Version 1.0. (http://www.boost.org/LICENSE_1_0.txt)
+# Port by Steven G. Johnson, also under Boost Software License, Version 1.0.
+#
+# Algorithm based on Mike Patefield and David Tandy, "Fast and accurate calculation
+# of Owen's T-function," Journal of Statistical Software, 5, pp. 1-25 (2000).
+
+using SpecialFunctions: erf, erfc
+using IrrationalConstants: inv2π, invsqrt2π
+
```
devmotion: These are already imported:

```suggestion
```

src/owens_t.jl
```diff
@@ -0,0 +1,247 @@
...
+        2, 3, 4, 4, 6, 6, 8, 8, 17, 17, 17, 17, 17, 12, 12,
+        2, 3, 4, 4, 6, 6, 18, 18, 18, 18, 17, 17, 17, 12, 12,
+    ) # 1-based indices (in 8x15 "matrix")
+    global owens_t_compute_code
+    function owens_t_compute_code(h::Real, a::Real)
+        ihint = something(findfirst(>=(h), hrange), length(hrange) + 1)
+        iaint = something(findfirst(>=(a), arange), length(arange) + 1)
+        return select[(iaint - 1) * 15 + ihint]
+    end
+end
```
devmotion: This code style seems a bit surprising and inconsistent with the rest of the package. Why not just defining `owens_t_compute_code` at the top level?

stevengj: So that `hrange`, `arange`, and `select` are not put in the global namespace.

stevengj: Alternatively, I could put the whole file into an `OwensT` submodule if you want, and get rid of all of the `owens_t_` prefixes.

src/owens_t.jl
stevengj: Done.

src/owens_t.jl
devmotion: I think it's fine to define global constants - that's already done in e.g. tvpack.jl

src/owens_t.jl
```diff
@@ -0,0 +1,244 @@
...
+Returns Owen's T function
+
+```math
+T(h,a) = \\frac{1}{2\\pi} \\int_0^a \\frac{e^{-h^2(1+x^2)/2}{1 + x^2} dx
+```
+
+(This is the probability of ``X > h`` and ``0 < Y < aX``, where ``X`` and ``Y`` are
+i.i.d. standard normal random variables.)
+"""
+owens_t(h::Real, a::Real) = _owens_t(promote(float(h), float(a))...)
```
devmotion: To avoid promoting e.g. `::Int, ::Float32` to `::Float64`:

```suggestion
owens_t(h::Real, a::Real) = _owens_t(map(float, promote(h, a))...)
```

src/owens_t.jl
stevengj: Okay, done.

stevengj:
CI failure seems unrelated:
```

in expression starting at D:\a\StatsFuns.jl\StatsFuns.jl\test\chainrules.jl:6
in expression starting at D:\a\StatsFuns.jl\StatsFuns.jl\test\runtests.jl:3
non-integer x = 1.100000non-integer x = 1.100000convergence failed in 'qtukey'
```
But that reminds me that it is probably worth adding a chain rule for `owens_t`, since there are simple formulas for the derivatives with respect to both arguments.

ext/StatsFunsChainRulesCoreExt.jl
```diff
@@ -82,4 +83,13 @@ ChainRulesCore.@scalar_rule(
     ),
 )
 
+ChainRulesCore.@scalar_rule(
+    owens_t(h::Real, a::Real),
+    @setup(),
```
devmotion: IIRC this can be omitted if not needed:
```suggestion
```

devmotion: Alternatively, you could use it to precompute `h^2` and use it in both derivatives.

stevengj:
Again, unrelated CI failure?
```
RNG of the outermost testset: Xoshiro(0x2ef1b0668e912b14, 0x91795677434389ce, 0xb8c8de0393cd7263, 0xb2f189b4e6006935, 0x63e4cdd08561e4e0)
ERROR: LoadError: Some tests did not pass: 109 passed, 0 failed, 1 errored, 0 broken.
in expression starting at /home/runner/work/StatsFuns.jl/StatsFuns.jl/test/chainrules.jl:6
in expression starting at /home/runner/work/StatsFuns.jl/StatsFuns.jl/test/runtests.jl:3
non-integer x = 1.100000non-integer x = 1.100000convergence failed in 'qtukey
non-integer x = -2.300000non-integer x = 1.200000non-integer x = 5.400000non-integer x = 11.900000non-integer x = -2.300000non-integer x = 1.200000non-integer x = 5.400000non-integer x = 11.900000non-integer x = -2.300000non-integer x = 1.200000non-integer x = 5.400000non-integer x = 11.900000non-integer x = -2.300000non-integer x = 1.200000non-integer x = 5.400000non-integer x = 11.900000non-integer x = -2.500000non-integer x = 1.300000non-integer x = 8.700000non-integer x = -2.500000non-integer x = 1.300000non-integer x = 8.700000non-integer x = -2.500000non-integer x = 1.300000non-integer x = 8.700000non-integer x = -2.500000non-integer x = 1.300000non-integer x = 8.700000Package StatsFuns errored during testing
Error: Process completed with exit code 1.
```
It seems like one of the tests has some non-negligible probability of failing?

ext/StatsFunsChainRulesCoreExt.jl
stevengj: Presumably the compiler's [CSE](https://en.wikipedia.org/wiki/Common_subexpression_elimination) pass will pull out expressions like `h^2` automatically.

devmotion:
That seems to indicate that some of the partials/tangents are generated incorrectly (or at least in a numerically problematic way) in the ChainRules tests, or some of the derivative rules have numerically problematic regimens (from the posted snipped it's unclear whether the rule or finite differencing would be to blame).

stevengj:
(To be clear, the occasional test failures are for other functions, unrelated to `owens_t` introduced here.)

devmotion:
> Unfortunately, the Boost tests don't seem to cover all of the branches, so I'll need to add more.

Do you plan to add more tests, or is this PR ready for review?

stevengj:
I've added tests to cover the branches that the Boost tests missed.   I think it is ready for review?

devmotion:
I was curious and checked the accuracy against Arb:

The reference is Arb's verified numerical integration of the integral representation

> T(h, a) = 1/(2π) ∫₀ᵃ exp(-h²(1+x²)/2)/(1+x²) dx

**Setup.** 518 `(h, a)` points covering all dispatch paths (T1–T6, the `a > 1` remapping, `a = 1`, `h = 0`, and a few sign-symmetric points), with `h` up to 20 and `a` up to 100. References were computed at 512-bit precision; the largest relative enclosure radius was `3.9e-31` (`eps(Float64) = 2.2e-16`), so the references are reliable.

**Float64 results.**
- relative error: max `1.49e-14`, mean `4.28e-16`, median `1.53e-16`
- ULP error: max `118`, mean `2.8`, median `1`
- the largest errors occur at `h = 20`, in the `a = 1` and `a > 1` branches
- the series methods T1–T6 stay within 9 ULPs

Per dispatch path:

```
path              count   max relerr  max ULPs
T1                  173     7.78e-16         7
T2                   30     6.42e-16         3
T3                   30     6.67e-16         5
T4                  108     1.51e-15         9
T5                   32     4.83e-16         3
T6                    8     4.02e-16         3
a > 1, h <= 0.67     48     3.99e-16         2
a > 1, h > 0.67      60     1.49e-14       118
h = 0                 2     0.00e+00         0
|a| = 1              27     1.49e-14       118
```

**Float32 / Float16.** These call the `Float64` path and round, so their relative error against Arb is at the level of `eps(Float32) ≈ 1e-7` and `eps(Float16) ≈ 1e-3`:

```
   h        a  Float32 relerr  Float16 relerr
0.25      0.5        9.62e-09        4.85e-05
 1.3      0.8        2.41e-08        7.85e-05
   2     0.99        4.82e-09        8.35e-05
 0.5        2        5.01e-08        1.48e-04
```

<details>
<summary>Script</summary>

```julia
# Accuracy check of `StatsFuns.owens_t` against Arb.
#
# Standalone script: not part of runtests.jl, and Arblib is not a test dependency.
# Run it in a temporary environment that has StatsFuns and Arblib:
#
#     julia> import Pkg; Pkg.activate(; temp = true)
#     julia> Pkg.develop(path = dirname(@__DIR__)); Pkg.add("Arblib")
#     julia> include("test/owens_t_error_analysis.jl")
#
# The reference is Arb's verified integration of the integral representation
#     T(h, a) = 1/(2π) ∫₀ᵃ exp(-h²(1+x²)/2)/(1+x²) dx
# The (h, a) grid covers every dispatch path (T1–T6 and the a > 1 remapping).
# Float32/Float16 just call the Float64 path and round, so they are only spot-checked.

using StatsFuns, Arblib, Printf, Statistics, Test

# 512-bit precision and a large evaluation budget keep the enclosure tight even for
# large h, where the integrand is a sharp spike near x = 0 and T(h, a) is as small
# as ~1e-89. Returns the value and the enclosure radius relative to it.
const INT_OPTS = Arblib.calc_integrate_opt_struct(0, 10_000_000, 10_000, 0, 0)

function owens_t_arb(h, a; prec = 512)
    H = Arb(h; prec)
    f = x -> exp(-H^2 * (1 + x^2) / 2) / (1 + x^2)
    I = Arblib.integrate(f, 0, a; prec, rtol = 1e-30, atol = Arblib.Mag(), opts = INT_OPTS)
    T = real(I) / (2 * Arb(π; prec))
    value = Float64(T)
    return value, Float64(Arblib.radius(Arf, T)) / abs(value)
end

# `meth` is local to `owens_t_dispatch` in src/owens_t.jl; mirror it to name which of
# T1–T6 handles a given (h, a) with 0 ≤ a ≤ 1.
const METH = (1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 3, 4, 4, 4, 4, 5, 6)

function regime_label(h, a)
    h == 0 && return "h = 0"
    abs(a) == 1 && return "|a| = 1"
    abs(a) > 1 && return abs(h) <= 0.67 ? "a > 1, h <= 0.67" : "a > 1, h > 0.67"
    return "T$(METH[StatsFuns.owens_t_compute_code(abs(h), abs(a))])"
end

# h and a values straddle the internal breakpoints (owens_t_hrange / owens_t_arange)
# and push past a = 1 into both remapping branches.
const HS = (0.01, 0.02, 0.05, 0.08, 0.1, 0.125, 0.2, 0.26, 0.4, 0.5, 0.6, 0.67,
            0.7, 1.0, 1.6, 1.7, 2.0, 2.33, 2.4, 3.0, 3.36, 3.4, 4.0, 4.8, 6.0, 10.0, 20.0)
const AS = (0.01, 0.025, 0.05, 0.09, 0.12, 0.15, 0.25, 0.36, 0.45, 0.5, 0.7, 0.9,
            0.99, 0.99999, 1.0, 1.2, 2.0, 5.0, 100.0)

grid = [(h, a) for h in HS for a in AS]
append!(grid, [(0.0, 0.5), (0.0, 2.0), (1.3, -0.4), (-1.3, 0.4), (-2.5, -0.8)])  # h = 0, sign symmetry

results = map(grid) do (h, a)
    computed = owens_t(h, a)
    ref, relrad = owens_t_arb(h, a)
    (; h, a, label = regime_label(h, a),
       relerr = abs(computed - ref) / abs(ref), ulps = abs(computed - ref) / eps(ref), relrad)
end

relerrs = [r.relerr for r in results]
ulpss = [r.ulps for r in results]

println("Owen's T Float64 vs Arb — $(length(results)) points")
@printf("relative error: max %.2e  mean %.2e  median %.2e\n", maximum(relerrs), mean(relerrs), median(relerrs))
@printf("ULP error:      max %.0f  mean %.1f  median %.0f\n", maximum(ulpss), mean(ulpss), median(ulpss))
@printf("max Arb enclosure relative radius: %.1e  (eps = %.1e)\n", maximum(r.relrad for r in results), eps(Float64))

println("\nper dispatch path:")
@printf("  %-16s %6s %12s %9s\n", "path", "count", "max relerr", "max ULPs")
for lab in sort(unique(r.label for r in results))
    sub = filter(r -> r.label == lab, results)
    @printf("  %-16s %6d %12.2e %9.0f\n", lab, length(sub), maximum(r.relerr for r in sub), maximum(r.ulps for r in sub))
end

println("\nworst 5 by relative error:")
@printf("  %8s %10s  %-16s %12s %9s\n", "h", "a", "path", "relerr", "ULPs")
for r in sort(results; by = r -> r.relerr, rev = true)[1:5]
    @printf("  %8.4g %10.5g  %-16s %12.2e %9.0f\n", r.h, r.a, r.label, r.relerr, r.ulps)
end

# Float32/Float16 round the inputs first, so take the reference at the rounded inputs
# to isolate the output error of the fallback from the input rounding.
println("\nFloat32 / Float16 spot check:")
@printf("  %8s %8s %14s %14s\n", "h", "a", "Float32 relerr", "Float16 relerr")
for (h, a) in ((0.25, 0.5), (1.3, 0.8), (2.0, 0.99), (0.5, 2.0))
    h32, a32 = Float32(h), Float32(a)
    h16, a16 = Float16(h), Float16(a)
    ref32, _ = owens_t_arb(Float64(h32), Float64(a32))
    ref16, _ = owens_t_arb(Float64(h16), Float64(a16))
    @printf("  %8.4g %8.4g %14.2e %14.2e\n", h, a,
            abs(owens_t(h32, a32) - ref32) / abs(ref32), abs(owens_t(h16, a16) - ref16) / abs(ref16))
end

@testset "owens_t accuracy vs Arb" begin
    @test maximum(relerrs) < 5e-14
    @test maximum(r.relrad for r in results) < 1e-20
end
```
</details>

src/owens_t.jl
```diff
@@ -0,0 +1,236 @@
...
+    1, 1, 2, 13, 13, 13, 13, 13, 13, 13, 13, 16, 16, 16, 9,
+    1, 2, 2, 3, 3, 5, 5, 14, 14, 15, 15, 16, 16, 16, 9,
+    2, 2, 3, 3, 3, 5, 5, 15, 15, 15, 15, 16, 16, 16, 10,
+    2, 2, 3, 5, 5, 5, 5, 7, 7, 16, 16, 16, 16, 16, 10,
+    2, 3, 3, 5, 5, 6, 6, 8, 8, 17, 17, 17, 12, 12, 11,
+    2, 3, 5, 5, 5, 6, 6, 8, 8, 17, 17, 17, 12, 12, 12,
+    2, 3, 4, 4, 6, 6, 8, 8, 17, 17, 17, 17, 17, 12, 12,
+    2, 3, 4, 4, 6, 6, 18, 18, 18, 18, 17, 17, 17, 12, 12,
+) # 1-based indices (in 8x15 "matrix")
+function owens_t_compute_code(h::Real, a::Real)
```
devmotion:
```suggestion
function owens_t_compute_code(h::Float64, a::Float64)
```

src/owens_t.jl
```diff
@@ -0,0 +1,236 @@
...
+    ihint = something(findfirst(>=(h), owens_t_hrange), length(owens_t_hrange) + 1)
+    iaint = something(findfirst(>=(a), owens_t_arange), length(owens_t_arange) + 1)
+    return owens_t_select[(iaint - 1) * 15 + ihint]
+end
+
+const owens_t_ord = (2, 3, 4, 5, 7, 10, 12, 18, 10, 20, 30, 0, 4, 7, 8, 20, 0, 0) # 53-bit precision table
+function owens_t_get_order(icode::Integer, ::Type{Float64})
```
devmotion:
```suggestion
function owens_t_get_order(icode::Int, ::Type{Float64})
```

src/owens_t.jl
```diff
@@ -0,0 +1,236 @@
...
+    return owens_t_ord[icode]
+end
+
+# compute the value of Owen's T function with method T1 from the reference paper
+function owens_t_T1(h::T, a::T, m::Integer) where {T <: Real}
```
devmotion: This seems neither accurate nor used with anything beyond `Float64`?

```suggestion
function owens_t_T1(h::Float64, a::Float64, m::Int)
```

src/owens_t.jl
```diff
@@ -0,0 +1,236 @@
...
+        jj += 2
+        aj *= as
+        dj = gj - dj
+        gj *= hs / j
+    end
+    return val
+end
+
+# compute the value of Owen's T function with method T2 from the reference paper
+function owens_t_T2(h::T, a::T, m::Integer, ah::T) where {T <: Real}
```
devmotion:
```suggestion
function owens_t_T2(h::Float64, a::Float64, m::Int, ah::Float64)
```

src/owens_t.jl
```diff
@@ -0,0 +1,236 @@
...
+        i += 1
+        zi = y * (ii * zi - vi)
+        vi *= as
+        ii += 2
+    end
+    return val * exp(-hs / 2) * invsqrt2π
+end
+
+# compute the value of Owen's T function with method T4 from the reference paper
+function owens_t_T4(h::T, a::T, m::Integer) where {T <: Real}
```
devmotion:
```suggestion
function owens_t_T4(h::Float64, a::Float64, m::Int)
```

src/owens_t.jl
```diff
@@ -0,0 +1,236 @@
...
+=#
+function owens_t_T5(h::Float64, a::Float64)
+    as = a * a
+    hs = -h * h / 2
+    r = 1 .+ as .* owens_t_pts
+    return sum(owens_t_wts .* exp.(hs .* r) ./ r) * a
+end
+
+# compute the value of Owen's T function with method T6 from the reference paper
+function owens_t_T6(h::T, a::T) where {T <: Float64}
```
devmotion:
```suggestion
function owens_t_T6(h::Float64, a::Float64)
```

src/owens_t.jl
```diff
@@ -0,0 +1,236 @@
...
+    return a < 0 ? -val : val # exploit that T(h,-a) == -T(h,a)
+end
+
+_owens_t(h::Float32, a::Float32) = Float32(owens_t(Float64(h), Float64(a)))
+_owens_t(h::Float16, a::Float16) = Float16(owens_t(Float64(h), Float64(a)))
+
+"""
+    owens_t(h::Real, a::Real)
+
+Returns Owen's T function
```
devmotion:
```suggestion
Return Owen's T function
```
Or maybe

```suggestion
Compute Owen's T function
```

src/owens_t.jl
```diff
@@ -0,0 +1,236 @@
...
+
+```math
+T(h,a) = \\frac{1}{2\\pi} \\int_0^a \\frac{e^{-h^2(1+x^2)/2}{1 + x^2} dx
```
devmotion:
```suggestion
T(h, a) = \\frac{1}{2\\pi} \\int_0^a \\frac{e^{-h^2(1+x^2)/2}}{1 + x^2} \, dx
```

src/owens_t.jl
```diff
@@ -0,0 +1,236 @@
...
+```
+
+(This is the probability of ``X > h`` and ``0 < Y < aX``, where ``X`` and ``Y`` are
+i.i.d. standard normal random variables.)
```
devmotion: Seems maybe a bit strange to put this in parentheses?

```suggestion
This is the probability of ``X > h`` and ``0 < Y < aX``, where ``X`` and ``Y`` are
i.i.d. standard normal random variables.
```

src/owens_t.jl
```diff
@@ -0,0 +1,236 @@
...
+"""
```
devmotion, line 234: Maybe add a link to the Wikipedia article? https://en.wikipedia.org/wiki/Owen%27s_T_function

devmotion:
One thing I also noticed: `owens_t(0, Inf)` returns `NaN` even though Owen's T function is 1/4 for these values. I think it would be good in general to add tests for such non-finite and special inputs (involving `Inf`, `-Inf`, `NaN`).

src/owens_t.jl
```diff
@@ -0,0 +1,236 @@
...
+    return a < 0 ? -val : val # exploit that T(h,-a) == -T(h,a)
+end
+
+_owens_t(h::Float32, a::Float32) = Float32(owens_t(Float64(h), Float64(a)))
+_owens_t(h::Float16, a::Float16) = Float16(owens_t(Float64(h), Float64(a)))
+
+"""
+    owens_t(h::Real, a::Real)
+
+Return Owen's T function
```
stevengj:
```suggestion
Return [Owen's T function](https://en.wikipedia.org/wiki/Owen%27s_T_function)
```

src/owens_t.jl
stevengj: Done.

stevengj:
I fixed the `Inf` and `NaN` cases and added some tests, including for the correct sign of `±0.0`.

src/owens_t.jl
```diff
@@ -0,0 +1,238 @@
...
+
+# compute the value of Owen's T function with method T2 from the reference paper
+function owens_t_T2(h::Float64, a::Float64, m::Int, ah::Float64)
+    maxii = m + m + one(m)
+    hs = h * h
+    as = -a * a
+    y = inv(hs)
+    ii = one(m)
+    val = zero(h)
+    vi = a * exp(-ah * ah / 2) * invsqrt2π
```
ararslan: StatsFuns has some handy distribution-related functions that would be nice to reuse for clarity and to benefit from any optimized methods that may exist.
```suggestion
    vi = a * normpdf(ah)
```

src/owens_t.jl
```diff
@@ -0,0 +1,238 @@
...
+    z = owens_t_znorm1(ah) / h
+    while true
+        val += z
+        maxii <= ii && break
+        z = y * (vi - ii * z)
+        vi *= as
+        ii += oftype(ii, 2)
+    end
+    return val * exp(-hs / 2) * invsqrt2π
```
ararslan:
```suggestion
    return val * normpdf(h)
```

src/owens_t.jl
```diff
@@ -0,0 +1,238 @@
...
+    0.89246305511006708555, -0.76893425990463999675, 0.5889352846848469325,
+    -0.38380345160440256652, 0.20317601701045299653, -0.82813631607004984866e-1,
+    0.24167984735759576523e-1, -0.44676566663971825242e-2, 0.39141169402373836468e-3,
+)
+function owens_t_T3(h::Float64, a::Float64, ah::Float64)
+    as = a * a
+    hs = h * h
+    y = inv(hs)
+    ii = one(h)
+    vi = a * exp(-ah * ah / 2) * invsqrt2π
```
ararslan:
```suggestion
    vi = a * normpdf(ah)
```

src/owens_t.jl
```diff
@@ -0,0 +1,238 @@
...
+    i = 1
+    while true
+        val += zi * owens_t_c2[i]
+        i == lastindex(owens_t_c2) && break
+        i += 1
+        zi = y * (ii * zi - vi)
+        vi *= as
+        ii += 2
+    end
+    return val * exp(-hs / 2) * invsqrt2π
```
ararslan:
```suggestion
    return val * normpdf(h)
```

src/owens_t.jl
```diff
@@ -0,0 +1,238 @@
...
+    # Use equation (2) in the paper to remap the arguments
+    # such that h>=0 and 0<=a<=1 for the call of the actual
+    # computation routine.
+    abs_a = abs(a)
+    abs_ah = abs_a * h
+
+    val = if abs_a <= 1
+        owens_t_dispatch(h, abs_a, abs_ah)
+    elseif !isfinite(abs_a)
+        isnan(abs_a) ? oftype(h, NaN) : erfc(h * invsqrt2) / 4
```
ararslan:
```suggestion
        isnan(abs_a) ? oftype(h, NaN) : normcdf(-h) / 2
```

ararslan (commented):
Drive-by minor review

ext/StatsFunsChainRulesCoreExt.jl
```diff
@@ -82,4 +83,12 @@ ChainRulesCore.@scalar_rule(
     ),
 )
 
+ChainRulesCore.@scalar_rule(
+    owens_t(h::Real, a::Real),
+    (
+        invsqrt2π * exp(-h^2 / 2) * erf((h * a) * invsqrt2) / -2,
```
devmotion: Could also reuse `normpdf` here

```suggestion
        normpdf(h) * erf((h * a) * invsqrt2) / -2,
```

src/owens_t.jl
```diff
@@ -0,0 +1,238 @@
...
+# Port by Steven G. Johnson, also under Boost Software License, Version 1.0.
+#
+# Algorithm based on Mike Patefield and David Tandy, "Fast and accurate calculation
+# of Owen's T-function," Journal of Statistical Software, 5, pp. 1-25 (2000).
+
+# P(-∞<Z<=x)-0.5 with Z being normally distributed.
+owens_t_znorm1(x::Real) = erf(x * invsqrt2) / 2
+
+# P(x<=Z<∞) with Z being normally distributed.
+owens_t_znorm2(x::Real) = erfc(x * invsqrt2) / 2
```
devmotion: We could also reuse `normccdf` here:

```suggestion
owens_t_znorm2(x::Real) = normccdf(x)
```

src/owens_t.jl
```diff
@@ -0,0 +1,238 @@
...
+    j, jj = one(m), oftype(h, 1)
+    while true
+        val += dj * aj / jj
+        m <= j && break
+        j += one(m)
+        jj += 2
+        aj *= as
+        dj = gj - dj
+        gj *= hs / j
+    end
```
devmotion, line 57: Maybe simpler to read?

```suggestion
    jj = oftype(h, 1)
    for j in 1:m
        val += dj * aj / jj
        jj += 2
        aj *= as
        dj = gj - dj
        gj *= hs / (j + 1)
    end
```

src/owens_t.jl
```diff
@@ -0,0 +1,238 @@
...
+    val = zero(h)
+    vi = a * normpdf(ah)
+    z = owens_t_znorm1(ah) / h
+    while true
+        val += z
+        maxii <= ii && break
+        z = y * (vi - ii * z)
+        vi *= as
+        ii += oftype(ii, 2)
+    end
```
devmotion, line 77: Maybe simpler?

```suggestion
    hs = h * h
    as = -a * a
    y = inv(hs)
    val = zero(h)
    vi = a * normpdf(ah)
    z = owens_t_znorm1(ah) / h
    for ii in 1:2:(2m + 1)
        val += z
        z = y * (vi - ii * z)
        vi *= as
    end
```

src/owens_t.jl
```diff
@@ -0,0 +1,238 @@
...
+    val = zero(h)
+    i = 1
+    while true
+        val += zi * owens_t_c2[i]
+        i == lastindex(owens_t_c2) && break
+        i += 1
+        zi = y * (ii * zi - vi)
+        vi *= as
+        ii += 2
+    end
```
devmotion, line 107: Maybe a bit simpler:

```suggestion
    for c in owens_t_c2
        val += zi * c
        zi = y * (ii * zi - vi)
        vi *= as
        ii += 2
    end
```

src/owens_t.jl
```diff
@@ -0,0 +1,238 @@
...
+    ai = a * exp(-hs * (1 - as) / 2) * inv2π
+    yi = one(h)
+    val = zero(h)
+    while true
+        val += ai * yi
+        maxii <= ii && break
+        ii += 2
+        yi = (1 - hs * yi) / ii
+        ai *= as
+    end
```
devmotion, line 126: Same here — iterate directly over the odd numbers?

```suggestion
    hs = h * h
    as = -a * a
    ai = a * exp(-hs * (1 - as) / 2) * inv2π
    yi = one(h)
    val = zero(h)
    for ii in 3:2:(2m + 3)
        val += ai * yi
        yi = (1 - hs * yi) / ii
        ai *= as
    end
```

src/owens_t.jl
```diff
@@ -0,0 +1,238 @@
...
+    # Handle some special cases first, these are from
+    # page 1077 of Owen's original paper:
+    iszero(h) && return atan(a) * inv2π
+    iszero(a) && return zero(h)
+    a == 1 && return owens_t_znorm2(-h) * owens_t_znorm2(h) / 2
+    @assert a <= 1 # when a>1 we call this routine with 1/a:
+
+    icode = owens_t_compute_code(h, a)
+    m = owens_t_get_order(icode, typeof(h))
+    meth = (1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 3, 4, 4, 4, 4, 5, 6) # 18 entries
```
devmotion, line 181: Should this also be a `const`?

src/owens_t.jl
stevengj: The loop in the original Boost code was written in this way in order to eliminate the unnecessary operations from the last loop iteration.   Probably not a big performance difference, but still — it's not purely stylistic.

src/owens_t.jl
stevengj: See [above](https://github.com/JuliaStats/StatsFuns.jl/pull/214#discussion_r3429604674)

src/owens_t.jl
stevengj: See [above](https://github.com/JuliaStats/StatsFuns.jl/pull/214#discussion_r3429604674)

src/owens_t.jl
stevengj: See [above](https://github.com/JuliaStats/StatsFuns.jl/pull/214#discussion_r3429604674)

src/owens_t.jl
stevengj: As far as I can tell LLVM generates identical code whether this is a local variable or a global `const`.

src/owens_t.jl
stevengj: Might as well just eliminate the entire `owens_t_znorm2` function.

src/owens_t.jl
devmotion: Ah true, I missed that 👍

devmotion (approved):
Looks good to me. If you update the version number, I could tag a new release right away when the PR is merged.

stevengj:
Done: bumped the version to 2.2.0

devmotion (approved):
Thank you!

PR #221, opened 2026-07-28
Avoid Int overflow in signrank and wilcox by normalising in place

### LINKED ISSUE #219: `signrankcdf` returns silently invalid probabilities from n = 72 (`Int` overflow in `signrankDP`)
Split out of the original report as suggested. The companion issue for `wilcox` is #220 — same root cause, different function and different threshold.

## Summary

`signrankDP` accumulates subset counts in `Int`. Those counts sum to `2^n` across the full support, so they exceed `typemax(Int)` and wrap around. `signrankcdf` then returns **silently invalid** values from **n = 72** onward — no error, no warning.

### Against the previous implementation

The `StatsFuns.RFunctions` submodule is gone in 2.2.0, so the comparison uses the standalone `Rmath` package — `Pkg.add("Rmath")`. (Rmath itself is still a StatsFuns dependency; it is the submodule that no longer exists.) The signed-rank null distribution is symmetric about `n(n+1)/4`, so the cdf evaluated there should be ≈ 0.5 for every `n`.

```julia
using StatsFuns, Rmath

for n in 69:74
    W = n * (n + 1) ÷ 4
    println(n, "\t", signrankcdf(n, W), "\t", Rmath.psignrank(W, n, true, false))
end
```

| n | `StatsFuns.signrankcdf` | `Rmath.psignrank` |
|---|---|---|
| 69 | 0.5000000000000004 | 0.49999999999999956 |
| 70 | 0.5000000000000001 | 0.5000000000000018 |
| 71 | 0.5011357204738435 | 0.5011357204738445 |
| **72** | **0.8604874010662797** | 0.5011124010662802 |
| **73** | **0.04492187500000003** | 0.49999999999999967 |
| **74** | **0.027343750000000014** | 0.5000000000000021 |

The two agree to ~1e-15 through `n = 71` and part company at **`n = 72`**. Note that the values do not drift — they swing above and below the correct answer, which is the signature of wraparound rather than of accumulated rounding.

### Self-contained checks

For completeness, the outputs are self-evidently not probabilities, with no reference needed:

```julia
using StatsFuns

# 1. the pmf does not sum to 1
for n in (50, 70, 71, 72, 80, 100)
    println(n, "\t", sum(signrankpdf(n, w) for w in 0:(n*(n+1)÷2)))
end
```

| n | `sum(signrankpdf)` | correct |
|---|---|---|
| 50 | 1.0 | 1 |
| 70 | 1.0 | 1 |
| 71 | 1.0 | 1 |
| **72** | **0.27734375** | 1 |
| **80** | **0.0006866455078124997** | 1 |
| **100** | **3.6379788070917093e-10** | 1 |

```julia
# 2. the cdf saturates at the midpoint of its support
for n in (80, 100, 120)
    println(n, "\tsupport 0:", n*(n+1)÷2, "\tcdf at midpoint = ", signrankcdf(n, n*(n+1)÷4))
end
```

| n | support | cdf at midpoint | correct |
|---|---|---|---|
| 80 | 0:3240 | 0.999654360398884 | ≈ 0.5 |
| 100 | 0:5050 | 0.9999999998209657 | ≈ 0.5 |
| 120 | 0:7260 | 1.0 | ≈ 0.5 |

## Cause

`signrankDP` builds `DP = zeros(Int, W + 1)` whose entries count subsets of `{1, …, n}`. Across the full support those counts sum to `2^n`, which `Int64` cannot represent beyond `n = 63`. `signrankcdf` sums only the smaller tail thanks to its symmetry reduction, which is why the observed onset is `n = 72` rather than 64.

`signrankccdf`, `signranklogcdf`, `signranklogccdf` and `signrankinvcdf` inherit the problem.

## Suggested fix

Accumulate in `Float64`. These counts are only ever consumed as ratios; `Float64` is exact to `2^53` and thereafter carries ~1e-16 relative error, immaterial for a cdf.

`Int128` or `BigInt` would also work, but cost more and only postpone the limit.

## Context

The pure-Julia implementations arrived in #173 and #174, released in #178, replacing the Rmath bindings compared against above.

Review on JuliaStats/HypothesisTests.jl#330 asked for tests comparing the new implementation against `Rmath.psignrank`. Those comparisons exist in `test/rmath.jl`, but only at `n = 4` and `n = 10`:

```julia
rmathcomp_tests(
    "signrank", [
        ((4), -2:12),
        ((4), -2.0:0.25:12.0),
        ((10), -2:57),
    ]
)
```

The only `n = 50` tests in that file are `cdf`/`invcdf` roundtrips of StatsFuns against itself, which pass regardless of whether the values are correct. So nothing in the suite would catch this.

The principal downstream consumer is HypothesisTests.jl, which imports `signrankcdf` and `signrankccdf` for the Wilcoxon signed-rank test.

## Versions

Julia 1.12.6 · StatsFuns 2.2.0 (also reproduced on 1.5.2) · Rmath 0.9.0

### LINKED ISSUE #220: `wilcoxcdf` returns silently invalid probabilities for large samples (`Int` overflow in `wilcox_partitions`)
Split out of #219 as suggested. That issue now covers `signrank` only; this one covers `wilcox`. Same root cause — combinatorial counts accumulated in `Int` — but a different function, a different threshold, and a different fix for the normaliser.

## Summary

`wilcox_partitions` accumulates rank-assignment counts in `Int`, and `wilcoxpdf` / `wilcoxcdf` normalise by `binomial(nx + ny, nx)`. Two consequences:

- **Silently invalid values** once the DP intermediates exceed `typemax(Int)`. This is the common case.
- **`OverflowError`** once `binomial(nx + ny, nx)` itself exceeds `typemax(Int)`, which happens later.

## Reproducer

The Mann–Whitney null distribution is symmetric about `nx·ny/2`, so the cdf evaluated at `nx·ny÷2` should be ≈ 0.5.

```julia
using StatsFuns, Rmath

for m in 30:34
    u = m * m ÷ 2
    sf = try wilcoxcdf(m, m, u) catch e sprint(showerror, e) end
    println(m, "\t", sf, "\t", Rmath.pwilcox(u, m, m, true, false))
end
```

| nx | ny | nx+ny | `binomial(nx+ny, nx)` | `StatsFuns.wilcoxcdf` | `Rmath.pwilcox` |
|---|---|---|---|---|---|
| 32 | 32 | 64 | 1832624140942590534 | 0.5026593344077379 | 0.5026593344077396 |
| **33** | **33** | 66 | 7219428434016265740 | **0.17794888610519205** | 0.5000000000000029 |
| **34** | **34** | 68 | 28453041475240576740 | **`OverflowError`** | 0.5024303008232744 |
| **30** | **36** | 66 | 5516694892996182896 | **0.7017246036833331** | 0.5025508900435341 |
| 20 | 48 | 68 | 82115378669464140 | 0.502662007134113 | 0.5026620071341129 |
| **20** | **60** | 80 | 3535316142212174320 | **0.5200602172127995** | 0.50219821409313 |
| 10 | 325 | 335 | 4283389682274562211 | 0.5006512415803868 | 0.5006512415803868 |
| **10** | **326** | 336 | 4414782003816726696 | **0.5086940261386015** | 0.500649276570064 |

Two things to note.

**The threshold is not `nx + ny`.** `nx = 20, ny = 48` is correct at `nx + ny = 68`, while `nx = 30, ny = 36` is wrong at `nx + ny = 66`. What matters is the number of rank assignments, `binomial(nx + ny, nx)`, which is largest for balanced splits. Very unbalanced groups stay correct much further out — `nx ≤ 9` is fine for every `ny` up to 400.

**The onset is well below `typemax(Int64)`** (`9223372036854775807`). Sweeping `nx` from 2 to 34 and finding the first failing `ny` in each case, the frontier sits between about `3.4e18` and `4.4e18`, i.e. roughly 40% of `typemax`. The Löffler recurrence subtracts as well as adds, so its intermediates overshoot the final counts and wrap before the total does.

The pmf confirms it independently, with no reference implementation needed:

```julia
julia> [sum(wilcoxpdf(m, m, u) for u in 0:(m*m)) for m in (30, 31, 32, 33)]
4-element Vector{Float64}:
 0.9999999999999998
 0.999999999999999
 1.0
 0.35589777221038377     # should be 1
```

## Cause

`wilcox_partitions` builds an `Int` table of partition counts bounded in total by `binomial(nx + ny, nx)`, with intermediates larger still. `wilcoxpdf` / `wilcoxcdf` then divide by `binomial(nx + ny, nx)`, which throws once that value alone exceeds `typemax(Int)` — `binomial(68, 34) ≈ 2.8e19`.

So there are two separate limits, and the *lower* one is the silent one.

`wilcoxccdf`, `wilcoxlogcdf`, `wilcoxlogccdf` and `wilcoxinvcdf` inherit both.

## Suggested fix

Accumulate the partition counts in `Float64`. They are only ever consumed as ratios; `Float64` is exact to `2^53` and thereafter carries ~1e-16 relative error, immaterial for a cdf.

For the normaliser, compute the binomial in log space (`logabsbinomial`) or normalise the partitions by their own sum, which avoids the throw and is one fewer thing to keep consistent.

`Int128` or `BigInt` would also work, but cost more and only postpone the limit.

## Context

The pure-Julia implementations arrived in #173 and #174, released in #178, replacing the Rmath bindings compared against above.

The Rmath comparisons in `test/rmath.jl` cover only very small groups, so nothing in the suite would catch this:

```julia
rmathcomp_tests(
    "wilcox", [
        ((3, 4), -2:13),
        ((3, 4), -2.0:13.0),
        ((4, 3), -2:13),
        ((4, 4), -2:17),
        ((1, 7), -2:8),
        ((1, 7), -2:8),
    ]
)
```

The `(10, 10)` tests alongside them are `cdf`/`invcdf` roundtrips of StatsFuns against itself, which pass regardless of whether the values are correct.

The principal downstream consumer is HypothesisTests.jl, which imports `wilcoxcdf` and `wilcoxccdf` for the Mann–Whitney U test.

## Versions

Julia 1.12.6 · StatsFuns 2.2.0 · Rmath 0.9.0

andreasnoack:
Fixes #219 and fixes #220.

`signrankDP` and `wilcox_partitions` accumulated combinatorial counts in `Int` and divided by the total count only at the end. Those counts grow like `2^n` and `binomial(nx + ny, nx)`, so they exceed `typemax(Int)` at moderate sample sizes: `signrankcdf` returned silently invalid probabilities from `n = 72` (#219), `wilcoxcdf` returned silently invalid probabilities from `binomial(nx + ny, nx) ≈ 3.4e18` (#220), and threw an `OverflowError` from `nx + ny = 68` (#219).

## Approach

Both recursions are linear and homogeneous, so the normaliser can be folded into the recursion rather than applied to its result. The counts are then never formed at all, and every intermediate value is a probability bounded by 1.

**`signrank`** halves at every step, `DP[i] = (DP[i] + DP[i + j]) / 2`, with a second pass over the entries whose target sum is smaller than `j` — those cannot contain `j`, so their counts are unchanged and their probabilities only pick up the factor `1/2`.

Halving is exact in binary floating point, and every count below `2^53` adds exactly, so the entries are bit-identical to the old ones divided by `2^n` while that holds, which covers `n <= 61`. Beyond that the additions round, but only barely — checked against exact `BigInt` counts:

```
n=60  max count=3.36e+15  every entry bit-identical to exact: true   max err=0.00 ulp
n=62  max count=1.28e+16  every entry bit-identical to exact: true   max err=0.50 ulp
n=70  max count=2.74e+18  every entry bit-identical to exact: false  max err=1.60 ulp
n=72  max count=1.05e+19  every entry bit-identical to exact: false  max err=2.12 ulp
```

The old implementation was valid up to `n = 71`, so in `62 <= n <= 71` results change, by at most about 1.6 ulp. For reference the old `sum(Base.Fix2(ldexp, -n) ∘ float, DP)` accumulation was itself 1.1e-15 to 1.8e-15 off at those sizes.

**`wilcox`** seeds the Löffler recurrence with `1 / binomial(nx + ny, nx)` instead of `1`, so it returns the probabilities directly. It is renamed `wilcox_probabilities` since it no longer returns partition counts. The seed is evaluated as a beta function via `logabsbinomial`, using `1 / binomial(n, k) = (n + 1) * beta(k + 1, n - k + 1)`, so it cannot overflow.

## Results

Maximum deviation from Rmath across the support:

```
n=71          max|Δcdf|=1.39e-15   max|Δpdf|=9.54e-18
n=200         max|Δcdf|=2.50e-15   max|Δpdf|=7.64e-18
nx=34 ny=34   max|Δcdf|=6.99e-15   max|Δpdf|=6.85e-17     (previously an OverflowError)
nx=150 ny=40  max|Δcdf|=1.31e-14   max|Δpdf|=1.78e-17
```

Every reproducer in the two issues:

```julia
julia> signrankcdf(72, 72 * 73 ÷ 4)   # was 0.8604874010662797
0.5011124010662795                     # Rmath: 0.5011124010662802

julia> wilcoxcdf(33, 33, 544)          # was 0.17794888610519205
0.5000000000000101                     # Rmath: 0.5000000000000029

julia> wilcoxcdf(30, 36, 540)          # was 0.7017246036833331
0.5025508900435314                     # Rmath: 0.5025508900435341

julia> wilcoxcdf(34, 34, 578)          # was OverflowError: binomial(68, 34) overflows
0.5024303008232822                     # Rmath: 0.5024303008232744

julia> sum(signrankpdf(100, w) for w in 0:5050)   # was 3.6379788070917093e-10
1.0
```

Sweeping #220's "frontier" — the first `ny` at which each `nx` disagrees with Rmath — finds no disagreement anywhere for `nx <= 34`, `ny <= 120`.

## Cost

The `wilcox` recurrence is 1.5x to 2.2x slower, growing with size (56.5 µs to 122 µs at `nx = ny = 33`): the old `Int` reduction vectorises because integer addition is associative, the `Float64` one cannot. Converting `sigmas` to `Vector{Float64}` makes no difference and `@simd` is slower, so this seems to be the price of leaving integer arithmetic. `signrank` is within 1.1x except for small `W`, where the added halving pass costs up to 1.5x.

One warning for future readers, left as a comment: deferring the `signrank` tail halvings to a single exact `ldexp` pass at the end looks 3.6x faster and is wrong, because the main loop would then add entries carrying different numbers of halvings. It returns 3.56e-01 where the answer is 8.08e-03.

## Inverse functions

`wilcoxinvcdf` was compared against `Rmath.qwilcox` at arguments that are themselves cdf values. The inverses locate a discontinuity by scanning for the first argument at which the cdf reaches `q`, so that comparison is decided by the last bit of the cdf: nudging the normaliser by a single ulp moves the result in either direction. It only passed before because the old code divided by an exactly representable `binomial`.

Rather than design the normaliser around it, `signrank` and `wilcox` are excluded from the Rmath inverse comparison, matching what #205 does for the other natively implemented discrete distributions, which round trip through their own cdf instead. Both already have such round trips in `test/rmath.jl` and they are unaffected. This also removes the existing `Sys.islinux()` special case, which was working around the same instability on one platform.

## Tests

Comparisons against Rmath at the distribution midpoint and in the left tail for `n` up to 200 and `nx + ny` up to 205, plus checks that the pdfs sum to one. The `wilcox` cases include the asymmetric ones from #220 — `(30, 36)`, `(20, 60)`, `(10, 326)` — which show that the silent threshold is set by `binomial(nx + ny, nx)` and not by `nx + ny`, since the Löffler recurrence subtracts as well as adds and its intermediates wrap before the counts do.

No compat bump is required: `logabsbinomial` is present in SpecialFunctions 1.8.1, the current lower bound.

## Scaling

The `wilcox` recurrence is O(U^2) in `U = nx * ny / 2` and `signrank` is O(n^3), both confirmed empirically (measured exponents 2.00 and 3.00). `signrank` is cheap enough to ignore, 0.14 s at `n = 1000`; `wilcox` costs 1.2 s at `nx = ny = 300` and about two minutes at `nx = ny = 1000`.

The seed `1 / binomial(nx + ny, nx)` is the probability of the least likely outcome, so it leaves the normal range long before the probabilities of interest do: subnormal from `nx + ny = 1030`, zero from `nx + ny = 1090`. Since the recurrence is homogeneous, a zero seed zeroes everything, so `wilcoxcdf` returned 0.776804 at `nx = ny = 540` and 0.0 from 545, where the answer is 0.5 by symmetry. The recurrence therefore works in units of `2^shift`, with `shift` chosen to keep the seed normal and removed from the scalar result with `ldexp`. `shift` is zero whenever the seed is already normal, so every input that worked before is bit for bit unchanged.

```
nx=ny=540   before: 0.776804   after: 0.5000389029915410   (normal approximation: 0.5000389192)
nx=ny=560   before: 0.000000   after: 0.5000368387803920   (normal approximation: 0.5000368536)
```

That moves the limit to about `nx + ny = 2050`, where the scaled values would overflow instead, and reaching it takes minutes of compute, so runtime rather than representability is now the binding constraint.

Two alternatives were measured and rejected. Switching to the normal approximation above some size is not viable: the absolute error only decays like `1/n` and the relative error is 28% at `cdf = 1e-6` even at `nx = ny = 140`, which is useless for a p-value. Switching to the Gaussian binomial product form, `prod (1 - q^(n+i)) / (1 - q^i)`, is 129x faster at `nx = ny = 400` and agrees to 1e-15 at small sizes, but it is numerically unstable: the numerator factors subtract nearly equal quantities, coefficients that must be non-negative reach -4.7e-13 by `nx = ny = 400` and -7.1e+113 by 700. It needs exact integer arithmetic, which is what we left to avoid the overflow in the first place.

One wart, noted for completeness: `ldexp` flushes the very last subnormal, so `wilcoxpdf(540, 540, 0)` is 0.0 where the exactly rounded value is 4.94e-324, the smallest representable number. That is the least likely outcome at a size whose every other value was previously garbage.

### DIFF
--- a/src/distrs/wilcox.jl
+++ b/src/distrs/wilcox.jl
@@ -66,13 +66,22 @@ function wilcox_probabilities(nx::Int, ny::Int, U::Int)
         end
     end
 
-    # Recursively compute pₘ,ₙ(a) / binomial(nx + ny, nx) for 0 <= a <= U.
-    # The seed is evaluated with `logabsbinomial` since `binomial(nx + ny, nx)` itself
-    # overflows `Int` from `nx + ny = 68` onwards. `logabsbinomial` evaluates the coefficient
-    # as a beta function, `1 / binomial(n, k) = (n + 1) * beta(k + 1, n - k + 1)`,
-    # and hence cannot overflow.
+    # Recursively compute pₘ,ₙ(a) / binomial(nx + ny, nx) for 0 <= a <= U, in units of
+    # 2^shift. The coefficient is evaluated with `logabsbinomial`, since `binomial(nx + ny,
+    # nx)` itself overflows `Int` from `nx + ny = 68` onwards, whereas `logabsbinomial`
+    # evaluates it as a beta function, `1 / binomial(n, k) = (n + 1) * beta(k + 1, n - k + 1)`.
+    #
+    # The seed is the probability of the least likely outcome, so it leaves the normal range
+    # long before the probabilities of interest do: it is subnormal from `nx + ny = 1030` and
+    # zero from `nx + ny = 1090`, and since the recurrence is homogeneous a seed of zero
+    # would silently zero every probability. Working in units of 2^shift avoids that. The
+    # recurrence is homogeneous so the choice of unit cannot affect the result, `ldexp`
+    # removes it from the scalar result exactly, and `shift` is zero whenever the seed is
+    # normal, which leaves every input that already worked bit for bit unchanged.
+    logbinom = first(logabsbinomial(nx + ny, nx))
+    shift = max(0, ceil(Int, (logbinom - 700) / logtwo))
     probabilities = Vector{Float64}(undef, U + 1)
-    probabilities[1] = exp(-first(logabsbinomial(nx + ny, nx)))
+    probabilities[1] = exp(shift * logtwo - logbinom)
     for a in 1:U
         p = 0.0
         for i in 1:a
@@ -81,7 +90,7 @@ function wilcox_probabilities(nx::Int, ny::Int, U::Int)
         probabilities[a + 1] = p / a
     end
 
-    return probabilities
+    return probabilities, shift
 end
 
 function wilcoxpdf(nx::Int, ny::Int, U::Float64)
@@ -93,8 +102,8 @@ function wilcoxpdf(nx::Int, ny::Int, U::Int)
         return 0.0
     end
     U = min(U, max_U - U)
-    probabilities = wilcox_probabilities(nx, ny, U)
-    return probabilities[end]
+    probabilities, shift = wilcox_probabilities(nx, ny, U)
+    return ldexp(probabilities[end], -shift)
 end
 
 function wilcoxlogpdf(nx::Int, ny::Int, U::Union{Float64, Int})
@@ -112,8 +121,8 @@ function wilcoxcdf(nx::Int, ny::Int, U::Int)
         return 1.0
     else
         U2 = max_U - U - 1
-        probabilities = wilcox_probabilities(nx, ny, min(U, U2))
-        p = sum(probabilities)
+        probabilities, shift = wilcox_probabilities(nx, ny, min(U, U2))
+        p = ldexp(sum(probabilities), -shift)
         return U2 < U ? 1.0 - p : p
     end
 end
--- a/src/distrs/wilcox.jl
+++ b/src/distrs/wilcox.jl
@@ -34,16 +34,6 @@ H. B. Mann, D. R. Whitney. "On a Test of Whether one of Two Random Variables is 
 A. Löffler: "Über eine Partition der nat. Zahlen und ihre Anwendung beim U-Test." Wissenschaftliche Zeitschrift der Martin-Luther-Universität Halle-Wittenberg; Mathematisch-Naturwissenschaftliche Reihe, XXXII'83 M, Heft 5, 87–89; available as https://upload.wikimedia.org/wikipedia/commons/f/f5/LoefflerWilcoxonMannWhitneyTest.pdf
 =#
 
-# `1 / binomial(n, k)`, without ever forming a coefficient that overflows `Int`.
-# `binomial(n, k)` is exact and converts to `Float64` exactly while it stays below `2^53`,
-# so it is used whenever that is guaranteed (`exp(36.7) < 2^53`). Above that we fall back to
-# `logabsbinomial`, which evaluates the coefficient as a beta function,
-# `1 / binomial(n, k) = (n + 1) * beta(k + 1, n - k + 1)`, and hence never overflows.
-function inv_binomial(n::Int, k::Int)
-    logbinom = first(logabsbinomial(n, k))
-    return logbinom < 36.7 ? inv(Float64(binomial(n, k))) : exp(-logbinom)
-end
-
 #=
 The recurrence below is linear and homogeneous in `pₘ,ₙ`, so seeding it with
 `1 / binomial(nx + ny, nx)` instead of `1` yields the probabilities directly.
@@ -76,9 +66,13 @@ function wilcox_probabilities(nx::Int, ny::Int, U::Int)
         end
     end
 
-    # Recursively compute pₘ,ₙ(a) / binomial(nx + ny, nx) for 0 <= a <= U
+    # Recursively compute pₘ,ₙ(a) / binomial(nx + ny, nx) for 0 <= a <= U.
+    # The seed is evaluated with `logabsbinomial` since `binomial(nx + ny, nx)` itself
+    # overflows `Int` from `nx + ny = 68` onwards. `logabsbinomial` evaluates the coefficient
+    # as a beta function, `1 / binomial(n, k) = (n + 1) * beta(k + 1, n - k + 1)`,
+    # and hence cannot overflow.
     probabilities = Vector{Float64}(undef, U + 1)
-    probabilities[1] = inv_binomial(nx + ny, nx)
+    probabilities[1] = exp(-first(logabsbinomial(nx + ny, nx)))
     for a in 1:U
         p = 0.0
         for i in 1:a
--- a/src/distrs/wilcox.jl
+++ b/src/distrs/wilcox.jl
@@ -34,14 +34,30 @@ H. B. Mann, D. R. Whitney. "On a Test of Whether one of Two Random Variables is 
 A. Löffler: "Über eine Partition der nat. Zahlen und ihre Anwendung beim U-Test." Wissenschaftliche Zeitschrift der Martin-Luther-Universität Halle-Wittenberg; Mathematisch-Naturwissenschaftliche Reihe, XXXII'83 M, Heft 5, 87–89; available as https://upload.wikimedia.org/wikipedia/commons/f/f5/LoefflerWilcoxonMannWhitneyTest.pdf
 =#
 
-@inline function wilcox_partitions(nx::Int, ny::Int, U::Int)
+# `1 / binomial(n, k)`, without ever forming a coefficient that overflows `Int`.
+# `binomial(n, k)` is exact and converts to `Float64` exactly while it stays below `2^53`,
+# so it is used whenever that is guaranteed (`exp(36.7) < 2^53`). Above that we fall back to
+# `logabsbinomial`, which evaluates the coefficient as a beta function,
+# `1 / binomial(n, k) = (n + 1) * beta(k + 1, n - k + 1)`, and hence never overflows.
+function inv_binomial(n::Int, k::Int)
+    logbinom = first(logabsbinomial(n, k))
+    return logbinom < 36.7 ? inv(Float64(binomial(n, k))) : exp(-logbinom)
+end
+
+#=
+The recurrence below is linear and homogeneous in `pₘ,ₙ`, so seeding it with
+`1 / binomial(nx + ny, nx)` instead of `1` yields the probabilities directly.
+This matters because the counts themselves grow like `binomial(nx + ny, nx)` and overflow
+`Int` from `nx + ny = 68` onwards, whereas the probabilities are bounded by 1.
+=#
+function wilcox_probabilities(nx::Int, ny::Int, U::Int)
     # This internal function expects 0 <= U <= nx * ny / 2
     if !(0 <= U <= (nx * ny) / 2)
-        throw(ArgumentError("`wilcox_partitions(nx, ny, U)` is only implemented for 0 <= U <= (nx * ny) / 2"))
+        throw(ArgumentError("`wilcox_probabilities(nx, ny, U)` is only implemented for 0 <= U <= (nx * ny) / 2"))
     end
 
-    # Due to symmetry, `wilcox_partitions(nx, ny, U) = wilcox_partitions(ny, nx, U)`
-    # Hence for simplicity we only consider the case `wilcox_partitions(min(nx, ny), max(nx, ny), U)` here
+    # Due to symmetry, `wilcox_probabilities(nx, ny, U) = wilcox_probabilities(ny, nx, U)`
+    # Hence for simplicity we only consider the case `wilcox_probabilities(min(nx, ny), max(nx, ny), U)` here
     m, n = minmax(nx, ny)
 
     # Compute σ(k) = ∑_{d|k} ϵ(d) d where
@@ -60,18 +76,18 @@ A. Löffler: "Über eine Partition der nat. Zahlen und ihre Anwendung beim U-Tes
         end
     end
 
-    # Recursively compute the number of partitions pₘ,ₙ(a) for 0 <= a <= U
-    partitions = Vector{Int}(undef, U + 1)
-    partitions[1] = 1
+    # Recursively compute pₘ,ₙ(a) / binomial(nx + ny, nx) for 0 <= a <= U
+    probabilities = Vector{Float64}(undef, U + 1)
+    probabilities[1] = inv_binomial(nx + ny, nx)
     for a in 1:U
-        p = 0
+        p = 0.0
         for i in 1:a
-            p += partitions[i] * sigmas[a + 1 - i]
+            p += probabilities[i] * sigmas[a + 1 - i]
         end
-        partitions[a + 1] = p ÷ a
+        probabilities[a + 1] = p / a
     end
 
-    return partitions
+    return probabilities
 end
 
 function wilcoxpdf(nx::Int, ny::Int, U::Float64)
@@ -83,8 +99,8 @@ function wilcoxpdf(nx::Int, ny::Int, U::Int)
         return 0.0
     end
     U = min(U, max_U - U)
-    partitions = wilcox_partitions(nx, ny, U)
-    return partitions[end] / binomial(nx + ny, nx)
+    probabilities = wilcox_probabilities(nx, ny, U)
+    return probabilities[end]
 end
 
 function wilcoxlogpdf(nx::Int, ny::Int, U::Union{Float64, Int})
@@ -102,8 +118,8 @@ function wilcoxcdf(nx::Int, ny::Int, U::Int)
         return 1.0
     else
         U2 = max_U - U - 1
-        partitions = wilcox_partitions(nx, ny, min(U, U2))
-        p = sum(float, partitions) / binomial(nx + ny, nx)
+        probabilities = wilcox_probabilities(nx, ny, min(U, U2))
+        p = sum(probabilities)
         return U2 < U ? 1.0 - p : p
     end
 end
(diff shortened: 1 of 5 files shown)


ArnoStrouwen (approved):

src/distrs/wilcox.jl
```diff
@@ -34,14 +34,30 @@ H. B. Mann, D. R. Whitney. "On a Test of Whether one of Two Random Variables is
...
 
-@inline function wilcox_partitions(nx::Int, ny::Int, U::Int)
+# `1 / binomial(n, k)`, without ever forming a coefficient that overflows `Int`.
+# `binomial(n, k)` is exact and converts to `Float64` exactly while it stays below `2^53`,
+# so it is used whenever that is guaranteed (`exp(36.7) < 2^53`). Above that we fall back to
+# `logabsbinomial`, which evaluates the coefficient as a beta function,
+# `1 / binomial(n, k) = (n + 1) * beta(k + 1, n - k + 1)`, and hence never overflows.
+function inv_binomial(n::Int, k::Int)
+    logbinom = first(logabsbinomial(n, k))
+    return logbinom < 36.7 ? inv(Float64(binomial(n, k))) : exp(-logbinom)
```
devmotion: It seems a bit inefficient to compute both `logabsbinomial` and `binomial` in the cases `logbinom < 36.7`? Is it actually necessary to fall back to using `binomial`? Or can we use an approximation of the binomial that guarantees that it doesn't overflow and branch depending on that, such that we only ever evaluate `binomial` or `logabsbinomial` but not both?

test/misc.jl
```diff
@@ -72,6 +72,22 @@ end
...
+
+    # The fast branch computes `binomial(n, k)` in `Int` and converts it to `Float64`,
+    # which is only exact while the coefficient stays below 2^53
+    @testset "the exact branch is only taken when it is exact" begin
+        @test all(
+            binomial(big(n), big(k)) <= 2^53
+                for n in 1:200 for k in 0:n
+                if first(SpecialFunctions.logabsbinomial(n, k)) < 36.7
+        )
+    end
```
devmotion: What is the point of this test? Just recomputing/ensuring that x < 36.7 implies exp(x) < 2^53?

andreasnoack: Ha. That seems like a pretty useless test. I'll remove it.

src/distrs/wilcox.jl
andreasnoack: The point is that `inv(Float64(binomial(n, k)))` is slightly more precise than `exp(-logbinom)` but I'm not sure it is worth the extra complexity. It should of course be possible to branch based on `n` and `k`, but that will be a more complicated branching rule.

devmotion: We could make use of

$$
\binom{n}{k} \leq 2^{n H(k/n)}
$$

where $H$ is the binary entropy function $H(p) = -p \log_2{p} - (1 - p) \log_2{(1 - p)}$, ie we could use `binomial` whenever $n H(k / n) < 53$.

Apparently, for $1 \leq k \leq n - 1$ the inequality could be tightened to

$$
\binom{n}{k} \leq \sqrt{\frac{n}{2\pi k (n - k)}} 2^{n H(k/n)}
$$

but that seems a bit uglier and more expensive to compute.

Ref https://en.wikipedia.org/wiki/Binomial_coefficient#Bounds_and_asymptotic_formulas

test/misc.jl
andreasnoack: Removed — it only restated the threshold it was handed. And the branch it was justifying is gone too, see below.

src/distrs/wilcox.jl
andreasnoack: You're right, and I've dropped the branch entirely — the normaliser is now always `exp(-first(logabsbinomial(nx + ny, nx)))`.

For the record on what the branch was buying, since it turned out not to be what I thought. On the pdf it is worth a factor of a few on an error that stays around 1e-15 either way, so it does not pay for itself there. What it did buy was the last bit of the cdf, and one thing depended on that: `wilcoxinvcdf` was compared against `Rmath.qwilcox` at arguments that are themselves cdf values, and since the inverse scans for the first `U` at which the cdf reaches `q`, that comparison is decided by the last bit. Dropping the branch produced 124 off-by-one failures.

But that is not a property worth designing around. Nudging the normaliser by a single ulp moves the result in either direction:

```
nx=1 ny=7  seed - 1ulp   invcdf BREAKS
nx=1 ny=7  seed (exact)  invcdf OK
nx=1 ny=7  seed + 1ulp   invcdf BREAKS
```

So instead I excluded `signrank` and `wilcox` from the Rmath inverse comparison, which is what #205 already does for the other natively implemented discrete distributions — they round trip through their own cdf rather than against the Rmath quantile functions. Both already have such round trips further down in `test/rmath.jl`, and they pass unchanged. That also let me delete the `Sys.islinux()` special case above, which was working around the same instability on one platform.

src/distrs/wilcox.jl
```diff
@@ -60,18 +66,22 @@ A. Löffler: "Über eine Partition der nat. Zahlen und ihre Anwendung beim U-Tes
         end
     end
 
-    # Recursively compute the number of partitions pₘ,ₙ(a) for 0 <= a <= U
-    partitions = Vector{Int}(undef, U + 1)
-    partitions[1] = 1
+    # Recursively compute pₘ,ₙ(a) / binomial(nx + ny, nx) for 0 <= a <= U.
+    # The seed is evaluated with `logabsbinomial` since `binomial(nx + ny, nx)` itself
+    # overflows `Int` from `nx + ny = 68` onwards. `logabsbinomial` evaluates the coefficient
+    # as a beta function, `1 / binomial(n, k) = (n + 1) * beta(k + 1, n - k + 1)`,
```
devmotion: This reads a bit strange - IIRC that's not how `logabsbinomial` generally evaluates `log(abs(binomial(...)))`? Otherwise, it would be also very surprising why we're not directly making use of this identity ourselves.

src/distrs/wilcox.jl
andreasnoack: I hadn't realized that Claude had replied on my behalf here.

src/distrs/wilcox.jl
andreasnoack: It has been updated so please have another look

src/distrs/wilcox.jl
```diff
@@ -60,18 +66,31 @@ A. Löffler: "Über eine Partition der nat. Zahlen und ihre Anwendung beim U-Tes
...
+    # nx)` itself overflows `Int` from `nx + ny = 68` onwards, whereas `logabsbinomial`
+    # evaluates it as a beta function, `1 / binomial(n, k) = (n + 1) * beta(k + 1, n - k + 1)`.
+    #
+    # The seed is the probability of the least likely outcome, so it leaves the normal range
+    # long before the probabilities of interest do: it is subnormal from `nx + ny = 1030` and
+    # zero from `nx + ny = 1090`, and since the recurrence is homogeneous a seed of zero
+    # would silently zero every probability. Working in units of 2^shift avoids that. The
+    # recurrence is homogeneous so the choice of unit cannot affect the result, `ldexp`
+    # removes it from the scalar result exactly, and `shift` is zero whenever the seed is
+    # normal, which leaves every input that already worked bit for bit unchanged.
```
devmotion: That's nonsense AFAICT - there's no bit for bit equality here since we use a different approach now, even in the case `shift = 0`.


```suggestion
    # normal.
```

devmotion (approved):

PR #227, opened 2026-07-31
Add `@inline` for `ForwardDiff.Dual` arguments

### LINKED ISSUE #223: pdf/logpdf/cdf functions are not inlined for `ForwardDiff.Dual` arguments
Many of the pdf/logpdf/cdf functions are not inlined when called with `ForwardDiff.Dual` arguments, although they are inlined for `Float64`. Adding `@inline` makes them 13-50% faster. Distributions.jl forwards to StatsFuns, so this affects AD users of `Normal`, `Gamma`, `Chisq` etc. as well, and there a call-site annotation is not an option.

Benchmark: loop over a varying `Float64` `x` with `Dual` parameters, `n = 4096`, min of 30 reps, normalised within each run by a loop calling the special function that dominates the body so the numbers do not depend on the machine. Medians over 4 runs of master and 2 with the patch. Same numbers at `-O2` and `-O3`. Julia 1.12.6, StatsFuns 2.2.1, ForwardDiff 1.4.3.

With `@inline` on all six functions in `norm.jl` (both the `z`-form and the 3-arg form), on `betalogpdf` and on `gammalogpdf`:

| | master | `@inline` | |
|---|---|---|---|
| `normpdf` | 3.43 | 1.73 | -50% |
| `normlogpdf` | 2.06 | 1.50 | -27% |
| `gammalogpdf` | 2.32 | 1.68 | -27% |
| `chisqlogpdf` | 2.34 | 1.79 | -24% |
| `binomlogpdf` | 6.68 | 5.15 | -23% |
| `binompdf` | 6.63 | 5.35 | -19% |
| `normlogcdf` | 0.95 | 0.77 | -18% |
| `normlogccdf` | 0.98 | 0.80 | -18% |
| `normcdf` | 0.79 | 0.66 | -17% |
| `normccdf` | 0.78 | 0.66 | -16% |
| `betapdf` | 4.66 | 4.07 | -13% |
| `betalogpdf` | 3.90 | 3.65 | -6% |

`gammapdf` +2%, `chisqpdf` +1%, `fdist*`, `tdist*` and `pois*` unchanged. Values are unchanged.

`chisqlogpdf` and `binomlogpdf` are not annotated themselves: they call `gammalogpdf` and `betalogpdf`, and get the improvement from those being inlined into their bodies, without their own call sites growing.

`chisqpdf` and `gammapdf` do not improve, in contrast to `chisqlogpdf` and `gammalogpdf`. For `gammapdf` the annotation only moves the calls to `_logabsgamma`, `_digamma` and `log` out of `gammalogpdf` and into `gammapdf` (`[gammalogpdf, exp]` becomes `[_logabsgamma, DomainError, _digamma, log, log, log, exp]`), so nothing is saved. `chisqpdf` is `gammapdf(k / 2, 2, x)` and just follows `gammapdf`.

The numbers are the same whether the parameters are loop-invariant or vary per element, so the improvements are not an artefact of hoisting.

I did benchmark `@inline` on the remaining functions too, but the code size is not free and there it does not pay. Summed over ten representative callers, LLVM lines emitted at `Dual` argument types:

| | annotations | improvements >8% | LLVM lines |
|---|---|---|---|
| master | - | - | 293 |
| `norm*`, `betalogpdf`, `gammalogpdf` | 16 | 11 | 1400 |
| every function that improves | 22 | 13 | 3093 |

Annotating `binomlogpdf` itself takes it from -23% to -31% but costs 926 LLVM lines per call site instead of 20, `fdistlogpdf` costs 466 and `tdistlogpdf` 361, for -10% and -12%. Not worth it IMO.

Two things worth recording:

- For the functions defined as a `promote` wrapper plus a `where {T}` worker, both methods need the annotation. Annotating only the worker moves the barrier rather than removing it: the worker is inlined into the wrapper, the wrapper then exceeds the threshold, and the call site is unchanged. That variant was slower than master for `poislogpdf` and `tdistlogpdf`.
- `pois` is left out because `@inline` does nothing for `poislogpdf` and costs `poispdf` about 12%.

(This issue was written with the assistance of generative AI.)

devmotion:
Fixes #223.

`@inline` on the six functions in `norm.jl` (both the `z`-form and the 3-arg form), on `betalogpdf` and on `gammalogpdf`. Without it these are not inlined for `ForwardDiff.Dual` arguments, although they are for `Float64`.

Relative cost of a loop over a varying `Float64` `x` with `Dual` parameters, normalised by a loop calling the special function that dominates the body, medians over 4 runs of master and 2 of this branch, same at `-O2` and `-O3`:

| | master | this PR | |
|---|---|---|---|
| `normpdf` | 3.43 | 1.73 | -50% |
| `normlogpdf` | 2.06 | 1.50 | -27% |
| `gammalogpdf` | 2.32 | 1.68 | -27% |
| `chisqlogpdf` | 2.34 | 1.79 | -24% |
| `binomlogpdf` | 6.68 | 5.15 | -23% |
| `binompdf` | 6.63 | 5.35 | -19% |
| `normlogcdf` | 0.95 | 0.77 | -18% |
| `normlogccdf` | 0.98 | 0.80 | -18% |
| `normcdf` | 0.79 | 0.66 | -17% |
| `normccdf` | 0.78 | 0.66 | -16% |
| `betapdf` | 4.66 | 4.07 | -13% |
| `betalogpdf` | 3.90 | 3.65 | -6% |

`chisqlogpdf`, `binomlogpdf`, `binompdf` and `betapdf` are not annotated themselves and improve because `gammalogpdf` and `betalogpdf` are inlined into their bodies, so their own call sites do not grow. Everything else is within ±8% and nothing regresses. Values are unchanged.

`fdistlogpdf`, `tdistlogpdf` and `poislogpdf` are deliberately left alone. See #223 for the numbers: annotating them costs 361-926 LLVM lines per call site for -10% to -12%, and for `poislogpdf` it gains nothing at all while costing `poispdf` about 12%. Annotating `binomlogpdf` on top of `betalogpdf` would take it from -23% to -31%, for 926 LLVM lines per call site instead of 20, which does not look worth it.

`test/inlining.jl` asserts the relative cost for `normpdf` and `normlogpdf`. Those are the two functions where the inlined and non-inlined versions are far enough apart to assert on: in the test's own loop the ratio to the baseline goes from 4.15 to 1.15 and from 1.91 to 1.08, against bounds of 2.0 and 1.4. For the remaining annotated functions the separation is only a factor 1.04-1.33, which is too small to bound reliably on shared CI runners. Removing the annotations from `norm.jl` makes both assertions fail.

(This PR was written with the assistance of generative AI.)

### DIFF
--- a/src/distrs/beta.jl
+++ b/src/distrs/beta.jl
@@ -5,8 +5,11 @@ using HypergeometricFunctions: _₂F₁
 # Julia implementations
 betapdf(α::Real, β::Real, x::Real) = exp(betalogpdf(α, β, x))
 
-betalogpdf(α::Real, β::Real, x::Real) = betalogpdf(promote(α, β, x)...)
-function betalogpdf(α::T, β::T, x::T) where {T <: Real}
+# `@inline` for `ForwardDiff.Dual` arguments, and so that callers such as
+# `binomlogpdf` and `betapdf` inline it into their bodies (#223). Both methods need
+# it, otherwise the un-annotated wrapper becomes the inlining barrier instead.
+@inline betalogpdf(α::Real, β::Real, x::Real) = betalogpdf(promote(α, β, x)...)
+@inline function betalogpdf(α::T, β::T, x::T) where {T <: Real}
     # we ensure that `log(x)` and `log1p(-x)` do not error
     y = clamp(x, 0, 1)
     val = xlogy(α - 1, y) + xlog1py(β - 1, -y) - logbeta(α, β)
--- a/src/distrs/gamma.jl
+++ b/src/distrs/gamma.jl
@@ -5,8 +5,11 @@ using HypergeometricFunctions: _₁F₁
 # Julia implementations
 gammapdf(k::Real, θ::Real, x::Real) = exp(gammalogpdf(k, θ, x))
 
-gammalogpdf(k::Real, θ::Real, x::Real) = gammalogpdf(promote(k, θ, x)...)
-function gammalogpdf(k::T, θ::T, x::T) where {T <: Real}
+# `@inline` for `ForwardDiff.Dual` arguments, and so that callers such as
+# `chisqlogpdf` inline it into their bodies (#223). Both methods need it,
+# otherwise the un-annotated wrapper becomes the inlining barrier instead.
+@inline gammalogpdf(k::Real, θ::Real, x::Real) = gammalogpdf(promote(k, θ, x)...)
+@inline function gammalogpdf(k::T, θ::T, x::T) where {T <: Real}
     # we ensure that `log(x)` does not error if `x < 0`
     xθ = max(x, 0) / θ
     val = -loggamma(k) - log(θ) - xθ
--- a/src/distrs/norm.jl
+++ b/src/distrs/norm.jl
@@ -1,5 +1,9 @@
 # functions related to normal distribution
 
+# The `@inline` annotations below are needed for `ForwardDiff.Dual` arguments: the
+# expansion of `exp`/`erfc`/`log` pushes these bodies over the inlining cost
+# threshold, so without them the functions are not inlined at their call sites (#223)
+
 function xval(μ::Real, σ::Real, z::Number)
     return if isinf(z) && iszero(σ)
         μ + one(σ) * z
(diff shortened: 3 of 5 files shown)


andreasnoack:
Duals can sometimes end up causing very big code blocks to be processed by LLVM. In particular, for nested Duals. Hence, I'm generally a bit reluctant to force inline functions. Might be worth analyzing why the heuristic doesn't inline smaller Duals.

Issue #228, opened 2026-08-17
Student-t CDF and quantile are significantly slower than Rmath


Santymax98:
While optimizing Student pair-copula conditionals in `VineCopulas.jl`, I found that the native Student-t CDF and especially the quantile can be substantially slower than the corresponding `Rmath` routines.

Using:

- StatsFuns v2.2.1
- Rmath v0.9.0

a small benchmark gives:

```julia
using BenchmarkTools, StatsFuns, Rmath

ν = 5.0
x = 1.5

@btime StatsFuns.tdistcdf($ν, $x)
# 355.423 ns (0 allocations)

@btime Rmath.pt($x, $ν)
# 230.705 ns (0 allocations)
````

For the quantile:

```julia
p = 0.95

@btime StatsFuns.tdistinvcdf($ν, $p)
# 1.396 μs (0 allocations)

@btime Rmath.qt($p, $ν)
# 561.773 ns (0 allocations)
```

and in the tail:

```julia
p = 1e-8

@btime StatsFuns.tdistinvcdf($ν, $p)
# 1.962 μs (0 allocations)

@btime Rmath.qt($p, $ν)
# 302.579 ns (0 allocations)
```

The numerical results agree essentially to machine precision.

For these examples, `Rmath` is about 1.5× faster for the CDF, 2.5× faster for a central quantile, and 6.5× faster for the tail quantile.

This matters downstream in `Copulas.jl` and `VineCopulas.jl`, where Student conditional CDFs and quantiles are repeatedly evaluated during Rosenblatt transforms and simulation. A related discussion is tracked in `lrnv/Copulas.jl#381`.

Would there be room to optimize the native `tdistcdf` / `tdistinvcdf` implementations, particularly the inverse CDF, while keeping the native Julia implementation?

andreasnoack:
Thanks for the report. I looked into where the time goes. StatsFuns has no dedicated Student-t code: `tdistcdf`/`tdistinvcdf` route through the F distribution into SpecialFunctions' `beta_inc`/`beta_inc_inv`, so essentially all of the time is spent in the incomplete beta function, and the quantile additionally pays for inverting it with a Newton iteration where every step evaluates the full `beta_inc`.

For the CDF, profiling turned up two fixable issues in SpecialFunctions' translation of the NSWC incomplete beta routines, now opened as PRs:

[...]

PR #229, opened 2026-08-17
Implement tdistinvcdf natively instead of inverting the incomplete beta

andreasnoack:
## Summary

Addresses the quantile part of #228 (see the task list in my comment there): replace the `tdistinvcdf` route through `fdistinvccdf`/`beta_inc_inv` — a Newton iteration where every step evaluates the full incomplete beta — with a pure Julia implementation based on VBA `invtdist` by Ian Smith:

- **Central region**: start from the Cornish–Fisher expansion of the t quantile in powers of 1/ν around the normal quantile (Fisher & Cornish, Technometrics 2 (1960)).
- **Tails**: start by inverting the leading power-law term `pr ≈ f(0)·√ν·|t|^(−ν)` with a closed-form next-order correction (crossover between the two starts follows Smith's empirical curves).
- **Polish**: Newton's method applied to `log(cdf)` — an ordinary Newton step near the center, nearly exact in the tails where the cdf is close to exponential in t. Each iteration costs one `beta_inc` evaluation; the pdf comes from the same argument reduction for one extra `pow`. Iterates are clamped to the half-line containing the solution.
- **Large ν**: for `ν ≥ 250` and `xn² ≤ ν/100` the Cornish–Fisher expansion has already converged to full precision and the polish is skipped (validated: relative error ≤ 2.2e-15 in this region, at the accuracy floor of `norminvcdf` itself).

Only `Float16` and `Float32` are routed explicitly through the Float64 kernel (per review); wider types such as `BigFloat` throw a `MethodError` rather than being silently computed at Float64 precision. The Newton polish stops without a verification-only cdf evaluation: after each step the remaining error is bounded by the quadratic-convergence estimate, with a 64x safety margin (per-df accuracy maxima re-validated unchanged).

## Accuracy

Against a 512-bit MPFR Newton reference on the incomplete beta representation:

- ν ∈ [1, 10⁴], p ∈ [1e-300, 1−1e-12] (250 points): max relative error **3.8e-14**, mean ~1e-15 — accuracy-equivalent to Rmath's `qt` (which shows up to 7.6e-9 at p = 1e-300 for small ν, and up to 5.5e-4 for ν < 1).
- Fixes genuine deep-tail failures of the current implementation: `tdistinvcdf(0.5, 1e-8)` returned `-Inf` (true value ≈ −1.03e15), and (ν = 2.5, p = 1e-12) was only good to 7e-9.

While testing round trips I also found that `tdistcdf` itself loses precision in deep tails (e.g. `tdistcdf(1, -3.18e7)` returns 9.49e-9 instead of 1.0e-8) because `fdistccdf` collapses the tail into `inv(1 + ν/x²) ≈ 1`. That is pre-existing and out of scope here, but the same paired argument reduction used by `_tdistcdf_pdf` in this PR would fix it — I can follow up.

## Performance

Apple Silicon, SpecialFunctions master (includes the merged but unreleased JuliaMath/SpecialFunctions.jl#542/#543):

| ν | p | this PR | master | Rmath `qt` |
|---|---|---|---|---|
| 5 | 0.3 | 353 ns | 680 ns | 283 ns |
| 5 | 0.95 | 445 ns | 801 ns | 370 ns |
| 5 | 1e-8 | 338 ns | 819 ns | 229 ns |
| 2.5 | 1e-12 | **233 ns** | 1613 ns | 274 ns |
| 50 | 0.999 | **365 ns** | 1117 ns | 365 ns |
| 0.5 | 1e-8 | **194 ns** | 8639 ns (wrong) | 5195 ns |
| 1000 | 0.975 | **106 ns** | 1454 ns | 189 ns |
| 1e5 | 0.999 | **108 ns** | 1433 ns | 185 ns |

With the released SpecialFunctions 2.8.3 the `beta_inc`-dependent rows are roughly 60-90 ns slower each.

## Test plan

- [x] Existing suite passes unchanged, including the Rmath comparisons (agreement with `qt` is within the 2e-14 tolerance on the test grids) and the `tdistinvcdf(0, 0.975) → NaN` regression test
- [x] New reference-value tests against the 512-bit reference, including deep tails, ν < 1, and the skip-polish region
- [x] Round trips `t → cdf → invcdf` (through the well-conditioned tail representation), edge cases (`p ∈ {0, 0.5, 1}`, `ν = Inf`, `NaN`s, out-of-range `p`), type stability for `Float16/32/64`

Disclosure: this PR was prepared by Claude Code at my direction; I have reviewed the implementation, the accuracy methodology, and the benchmarks.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/distrs/tdist.jl
+++ b/src/distrs/tdist.jl
@@ -97,84 +97,77 @@ function _tdistinvcdf(ν::Float64, p::Float64)
     logB = logbeta(ν / 2, 0.5)
     B = exp(logB)
 
-    local tp::Float64
-    tprob = 0.0
-    tpdif = 0.0
-    if pr >= 0.5 || (ν >= 1.0 && _tdist_cf_start_is_better(pr, ν))
+    tp = if pr >= 0.5 || (ν >= 1.0 && _tdist_cf_start_is_better(pr, ν))
         # Cornish-Fisher expansion of the t quantile in powers of 1/ν around the
         # normal quantile (Fisher & Cornish, Technometrics 2 (1960), 209-225)
         xn = norminvcdf(pr)
         x = xn * xn
-        tp = (((((27.0 * x + 339.0) * x + 930.0) * x - 1782.0) * x - 765.0) * x + 17955.0) / (368640.0 * ν)
-        tp = (tp + ((((79.0 * x + 776.0) * x + 1482.0) * x - 1920.0) * x - 945.0) / 92160.0) / ν
-        tp = (tp + (((3.0 * x + 19.0) * x + 17.0) * x - 15.0) / 384.0) / ν
-        tp = (tp + ((5.0 * x + 16.0) * x + 3.0) / 96.0) / ν
-        tp = (tp + (x + 1.0) / 4.0) / ν
-        tp = xn * (1.0 + tp)
+        t = (((((27.0 * x + 339.0) * x + 930.0) * x - 1782.0) * x - 765.0) * x + 17955.0) / (368640.0 * ν)
+        t = (t + ((((79.0 * x + 776.0) * x + 1482.0) * x - 1920.0) * x - 945.0) / 92160.0) / ν
+        t = (t + (((3.0 * x + 19.0) * x + 17.0) * x - 15.0) / 384.0) / ν
+        t = (t + ((5.0 * x + 16.0) * x + 3.0) / 96.0) / ν
+        t = (t + (x + 1.0) / 4.0) / ν
+        t = xn * (1.0 + t)
         # for large ν the expansion has already converged to full precision and the
         # Newton polish can be skipped (validated relative error <= 2.2e-15 in this
         # region against a high-precision reference)
         if ν >= 250.0 && x <= ν / 100.0
-            return p > 0.5 ? -tp : tp
+            return p > 0.5 ? -t : t
         end
-        tprob = 0.0
-        tpdif = 1.0 + abs(tp)
+        t
     elseif ν < 1.0
         # leading power-law term of the tail solved in log space:
         # pr ≈ (ν / t^2)^(ν / 2) / (ν * B)
         lν = log(ν)
-        tp = -exp(lν / 2 - (log(pr) + lν + logB) / ν)
-        isfinite(tp) || return p > 0.5 ? -tp : tp
-        tprob, f = _tdistcdf_pdf(ν, B, tp)
-        if f < floatmin(Float64)
-            tpdif = 0.0
-        else
-            tpdif = tprob / f * log1p((tprob - pr) / pr)
-            tpnew = tp - tpdif
-            tp = tpnew < 0.0 ? tpnew : tp / 2
-        end
+        -exp(lν / 2 - (log(pr) + lν + logB) / ν)
     else
-        # invert the leading power-law term of the tail: pr ≈ f(0) * √ν * |t|^(-ν)
-        # where f(0) = 1 / (√ν * B) is the density at zero, map back to the t scale,
-        # apply one closed-form next-order correction, and one Newton step
+        # invert the leading power-law term of the tail, pr ≈ f(0) * √ν * |t|^(-ν)
+        # where f(0) = 1 / (√ν * B) is the density at zero, map back to the t
+        # scale, and apply one closed-form next-order correction
         u = exp(-log(ν * B * pr) / ν)
-        tp = -sqrt(ν) * sqrt(u - 1.0) * sqrt(u + 1.0)
-        isfinite(tp) || return p > 0.5 ? -tp : tp
-        tpdif = tp / ν
-        tpdif = -log1p((0.5 - 1.0 / (ν + 2.0)) / (1.0 + tpdif * tp)) * (tpdif + 1.0 / tp)
-        tp -= tpdif
-        tprob, f = _tdistcdf_pdf(ν, B, tp)
-        if f < floatmin(Float64)
-            tpdif = 0.0
-        else
-            tpdif = tprob / f * log1p((tprob - pr) / pr)
-            tpnew = tp - tpdif
-            tp = tpnew < 0.0 ? tpnew : tp / 2
+        t = -sqrt(ν) * sqrt(u - 1.0) * sqrt(u + 1.0)
+        if isfinite(t)
+            d = t / ν
+            t -= -log1p((0.5 - 1.0 / (ν + 2.0)) / (1.0 + d * t)) * (d + 1.0 / t)
         end
+        t
     end
+    # the true quantile may overflow in the extreme tails
+    isfinite(tp) || return p > 0.5 ? -tp : tp
 
     # Newton iteration applied to log(cdf): step = cdf / pdf * log(cdf / pr).
     # Near the center this is an ordinary Newton step; in the tails, where the cdf
     # is nearly exponential in t, the log transform makes the problem nearly linear.
     iter = 0
-    while abs(tprob - pr) > smalllpr && abs(tpdif) > small * (1.0 + abs(tp))
-        (iter += 1) > 100 && break
+    while true
         tprob, f = _tdistcdf_pdf(ν, B, tp)
+        # the density underflows only where the start already carries full precision
         f < floatmin(Float64) && break
         tpdif = tprob / f * log1p((tprob - pr) / pr)
+        # second-order coefficient of the iteration, |d²log(cdf)/dt² / dlog(cdf)/dt|,
+        # for the quadratic-convergence estimate of the error remaining after the step
+        curv = abs(-(ν + 1.0) * tp / (ν + tp * tp) - f / tprob)
         tpnew = tp - tpdif
         # keep the iterate in the negative half-line, where the solution lies
         tp = tpnew < 0.0 ? tpnew : tp / 2
+        tol = small * (1.0 + abs(tp))
+        # converged if the cdf already matched, the step was negligible, or the
+        # estimated error remaining after the step is negligible (the exact
+        # second-order coefficient is 1/2; the factor 32 is a 64x safety margin)
+        abs(tprob - pr) <= smalllpr && break
+        abs(tpdif) <= tol && break
+        32.0 * curv * tpdif * tpdif <= tol && break
+        (iter += 1) >= 100 && break
     end
     return p > 0.5 ? -tp : tp
 end
 
-# The kernel operates in Float64, like the Rmath-based functions elsewhere in the
-# package; argument types with more precision than Float64 are truncated.
-function tdistinvcdf(ν::T, p::T) where {T <: Real}
-    return convert(float(T), _tdistinvcdf(Float64(ν), Float64(p)))
-end
-tdistinvcdf(ν::Real, p::Real) = tdistinvcdf(promote(ν, p)...)
+# Only Float16 and Float32 are routed explicitly through the Float64 kernel;
+# wider types such as BigFloat are unsupported rather than silently computed
+# at Float64 precision.
+_tdistinvcdf(ν::Float16, p::Float16) = convert(Float16, _tdistinvcdf(Float64(ν), Float64(p)))
+_tdistinvcdf(ν::Float32, p::Float32) = convert(Float32, _tdistinvcdf(Float64(ν), Float64(p)))
+tdistinvcdf(ν::Real, p::Real) = _tdistinvcdf(map(float, promote(ν, p))...)
 
 tdistinvccdf(ν::Real, p::Real) = -tdistinvcdf(ν, p)
 function tdistinvlogcdf(ν::T, logp::T) where {T <: Real}
--- a/src/distrs/tdist.jl
+++ b/src/distrs/tdist.jl
@@ -36,6 +36,142 @@ tdistlogcdf(ν::Real, x::Real) = tdistlogcdf(map(float, promote(ν, x))...)
 
 tdistlogccdf(ν::Real, x::Real) = tdistlogcdf(ν, -x)
 
+# Pure Julia implementation of the inverse CDF, based on VBA invtdist by Ian Smith:
+# start from a Cornish-Fisher expansion of the quantile around the normal quantile
+# (central region) or from inverting the leading power-law term of the tail (tail
+# region), then polish with Newton's method applied to log(cdf), where each
+# iteration costs a single evaluation of the incomplete beta function.
+
+# Crossover curves between the Cornish-Fisher and tail starting approximations.
+# Based on VBA BetterThanTailApprox by Ian Smith.
+function _tdist_cf_start_is_better(pr::Float64, ν::Float64)
+    if ν <= 2.0
+        return pr > 0.25 * exp((1.0 - ν) * 1.78514841051368)
+    elseif ν <= 5.0
+        return pr > 0.045 * exp((2.0 - ν) * 1.30400766847605)
+    elseif ν <= 20.0
+        return pr > 0.0009 * exp((5.0 - ν) * 0.921034037197618)
+    else
+        return pr > 9.0e-10 * exp((20.0 - ν) * 0.690775527898214)
+    end
+end
+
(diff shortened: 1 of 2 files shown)


src/distrs/tdist.jl
```diff
@@ -36,16 +36,145 @@ tdistlogcdf(ν::Real, x::Real) = tdistlogcdf(map(float, promote(ν, x))...)
...
+    return p > 0.5 ? -tp : tp
+end
+
+# The kernel operates in Float64, like the Rmath-based functions elsewhere in the
+# package; argument types with more precision than Float64 are truncated.
+function tdistinvcdf(ν::T, p::T) where {T <: Real}
+    return convert(float(T), _tdistinvcdf(Float64(ν), Float64(p)))
 end
-tdistinvcdf(ν::Real, p::Real) = tdistinvcdf(map(float, promote(ν, p))...)
+tdistinvcdf(ν::Real, p::Real) = tdistinvcdf(promote(ν, p)...)
```
devmotion: IMO that's worse than not supporting `BigFloat` as it indicates an incorrect precision of the result. IMO we should use the same approach as in e.g. SpecialFunctions and only explicitly route `Float16` and `Float32` through the `Float64`-specific method.

Something like

```suggestion
tdistinvcdf(ν::Real, p::Real) = _tdistinvcdf(map(float, promote(ν, p))...)
_tdistinvcdf(ν::Float16, p::Float16) = convert(Float16, _tdistinvcdf(Float64(ν), Float64(p)))
_tdistinvcdf(ν::Float32, p::Float32) = convert(Float32, _tdistinvcdf(Float64(ν), Float64(p)))
```

src/distrs/tdist.jl
```diff
@@ -36,16 +36,145 @@ tdistlogcdf(ν::Real, x::Real) = tdistlogcdf(map(float, promote(ν, x))...)
...
+
+    # work with the smaller tail; the result is negated for p > 1/2 on return
+    pr = p > 0.5 ? 1.0 - p : p
+    small = 1.0e-14
+    smalllpr = -small * log(pr) * pr
+
+    logB = logbeta(ν / 2, 0.5)
+    B = exp(logB)
+
+    local tp::Float64
```
devmotion: codecov says this line is not covered - seems wrong?

src/distrs/tdist.jl
```diff
@@ -36,16 +36,145 @@ tdistlogcdf(ν::Real, x::Real) = tdistlogcdf(map(float, promote(ν, x))...)
...
+        tpdif = 1.0 + abs(tp)
+    elseif ν < 1.0
+        # leading power-law term of the tail solved in log space:
+        # pr ≈ (ν / t^2)^(ν / 2) / (ν * B)
+        lν = log(ν)
+        tp = -exp(lν / 2 - (log(pr) + lν + logB) / ν)
+        isfinite(tp) || return p > 0.5 ? -tp : tp
+        tprob, f = _tdistcdf_pdf(ν, B, tp)
+        if f < floatmin(Float64)
+            tpdif = 0.0
```
devmotion: codecov says this line is not covered

src/distrs/tdist.jl
```diff
@@ -36,16 +36,145 @@ tdistlogcdf(ν::Real, x::Real) = tdistlogcdf(map(float, promote(ν, x))...)
...
+        # apply one closed-form next-order correction, and one Newton step
+        u = exp(-log(ν * B * pr) / ν)
+        tp = -sqrt(ν) * sqrt(u - 1.0) * sqrt(u + 1.0)
+        isfinite(tp) || return p > 0.5 ? -tp : tp
+        tpdif = tp / ν
+        tpdif = -log1p((0.5 - 1.0 / (ν + 2.0)) / (1.0 + tpdif * tp)) * (tpdif + 1.0 / tp)
+        tp -= tpdif
+        tprob, f = _tdistcdf_pdf(ν, B, tp)
+        if f < floatmin(Float64)
+            tpdif = 0.0
```
devmotion: codecov says this is not covered

devmotion:
The benchmark looks great, in particular with the latest commit 👍

src/distrs/tdist.jl
andreasnoack: Indeed, it must have been wrong. However, Claude changed the structure a bit with the latest commit so now the measured coverage is complete.

andreasnoack:
@devmotion do you have any additional comments?

src/distrs/tdist.jl
```diff
@@ -36,16 +36,138 @@ tdistlogcdf(ν::Real, x::Real) = tdistlogcdf(map(float, promote(ν, x))...)
...
+    isfinite(tp) || return p > 0.5 ? -tp : tp
+
+    # Newton iteration applied to log(cdf): step = cdf / pdf * log(cdf / pr).
+    # Near the center this is an ordinary Newton step; in the tails, where the cdf
+    # is nearly exponential in t, the log transform makes the problem nearly linear.
+    iter = 0
+    while true
+        tprob, f = _tdistcdf_pdf(ν, B, tp)
+        # the density underflows only where the start already carries full precision
+        f < floatmin(Float64) && break
```
devmotion, line 145: For `ν < 1` the exponent `(ν + 1) / 2` is smaller than 1, so `f` is much larger than `k2` and stays well above `floatmin` long after `k2` itself has become subnormal. In that window `beta_inc` only sees a handful of significant bits of its argument, `tprob` comes back wrong in the third digit, and the step throws away a start that was already good to ~3e-14. I get a relative error of 3.4e-1 for `(0.9, 1e-146)` and 1.7e-1 for `(0.35, 1e-57)`, and for eg `(0.5, 1e-81)` and `(0.9, 1e-145)` the loop runs into the iteration cap - 100 `beta_inc` evaluations, ~11 µs instead of ~200 ns. Per-df maxima over `p ∈ [1e-300, 1/2]`: 3.1e-8 at ν = 0.2, 1.7e-1 at 0.35, 1.5e-2 at 0.5, 8.9e-4 at 0.7, 3.4e-1 at 0.9. `qt` is fine (~5e-14) at all of these.

Could we guard on the incomplete beta argument rather than on the density, ie return `k2` from `_tdistcdf_pdf` and break on `k2 < floatmin(Float64)`? With that the maximum for ν < 1 is back to 1.6e-13. The comment above assumes the two underflow together, which only holds for `ν ≥ 1`.

src/distrs/tdist.jl
```diff
@@ -36,16 +36,138 @@ tdistlogcdf(ν::Real, x::Real) = tdistlogcdf(map(float, promote(ν, x))...)
...
+        tpdif = tprob / f * log1p((tprob - pr) / pr)
+        # second-order coefficient of the iteration, |d²log(cdf)/dt² / dlog(cdf)/dt|,
+        # for the quadratic-convergence estimate of the error remaining after the step
+        curv = abs(-(ν + 1.0) * tp / (ν + tp * tp) - f / tprob)
+        tpnew = tp - tpdif
+        # keep the iterate in the negative half-line, where the solution lies
+        tp = tpnew < 0.0 ? tpnew : tp / 2
+        tol = small * (1.0 + abs(tp))
```
devmotion, line 153: `tol` is essentially absolute for `|tp| ≪ 1`, and `(tprob - pr) / pr` cancels once `tprob ≈ pr ≈ 1/2`, so we lose relative accuracy close to the median. For ν = 1 I get 2.7e-13 at `p = 0.5 - 1e-4`, 2.6e-11 at `0.5 - 1e-6`, 4.6e-8 at `0.5 - 1e-10` and 4.2e-4 at `0.5 - 1e-14`, where the current implementation stays at ~1e-16 throughout (Rmath is as bad as this PR here). Same shape for every ν up to ~100. The absolute error is fine everywhere, so perhaps this is acceptable, but it is a regression and the description claims accuracy equivalence with `qt` without mentioning it.

It also looks cheap to avoid: `beta_inc` already returns `1/2 - cdf` as its second component, and `0.5 - pr` is exact for `pr ≥ 1/4`. Iterating on those two for `pr > 1/4` gives me ~1e-16 over the whole range with the same number of `beta_inc` evaluations.

src/distrs/tdist.jl
```diff
@@ -36,16 +36,138 @@ tdistlogcdf(ν::Real, x::Real) = tdistlogcdf(map(float, promote(ν, x))...)
...
+    elseif p == 0.5
+        return 0.0
+    end
+
+    # work with the smaller tail; the result is negated for p > 1/2 on return
+    pr = p > 0.5 ? 1.0 - p : p
+    small = 1.0e-14
+    smalllpr = -small * log(pr) * pr
+
+    logB = logbeta(ν / 2, 0.5)
```
devmotion, line 97: Neither `logB` nor `B` is used before the early return for large `ν` below, and together they cost ~65 ns. Could we move them into the branches that need them? That takes `tdistinvcdf(1000, 0.975)` and `tdistinvcdf(1e5, 0.999)` from ~105 ns to ~50 ns here.

src/distrs/tdist.jl
```diff
@@ -36,16 +36,138 @@ tdistlogcdf(ν::Real, x::Real) = tdistlogcdf(map(float, promote(ν, x))...)
...
+        end
+        t
+    elseif ν < 1.0
+        # leading power-law term of the tail solved in log space:
+        # pr ≈ (ν / t^2)^(ν / 2) / (ν * B)
+        lν = log(ν)
+        -exp(lν / 2 - (log(pr) + lν + logB) / ν)
     else
-        return sqrt(fdistinvccdf(one(ν), ν, 2 * (1 - p)))
+        # invert the leading power-law term of the tail, pr ≈ f(0) * √ν * |t|^(-ν)
```
devmotion, line 124: I don't quite follow this one: `f(0) * √ν * |t|^(-ν)` is `|t|^(-ν) / B`, and inverting that gives `(B * pr)^(-1/ν)` rather than `√ν * (ν * B * pr)^(-1/ν)`. AFAICT the relation actually inverted here is the same as in the branch above, `pr ≈ (ν / t^2)^(ν/2) / (ν * B)`, and the only difference is that `sqrt(u - 1) * sqrt(u + 1)` keeps `x = ν / (ν + t^2)` exact instead of dropping to the leading power of `t`. Could we use the same wording as above? The description has the same formula.

src/distrs/tdist.jl
```diff
@@ -36,16 +36,138 @@ tdistlogcdf(ν::Real, x::Real) = tdistlogcdf(map(float, promote(ν, x))...)
...
+
+    # work with the smaller tail; the result is negated for p > 1/2 on return
+    pr = p > 0.5 ? 1.0 - p : p
+    small = 1.0e-14
+    smalllpr = -small * log(pr) * pr
+
+    logB = logbeta(ν / 2, 0.5)
+    B = exp(logB)
+
+    tp = if pr >= 0.5 || (ν >= 1.0 && _tdist_cf_start_is_better(pr, ν))
```
devmotion, line 100: Is `pr >= 0.5` reachable? `p == 0.5` returns earlier and `1 - p` is exact for `p > 1/2`, so `pr < 0.5` always as far as I can tell.

src/distrs/tdist.jl
```diff
@@ -36,16 +36,138 @@ tdistlogcdf(ν::Real, x::Real) = tdistlogcdf(map(float, promote(ν, x))...)
...
+
+# Only Float16 and Float32 are routed explicitly through the Float64 kernel;
+# wider types such as BigFloat are unsupported rather than silently computed
+# at Float64 precision.
+_tdistinvcdf(ν::Float16, p::Float16) = convert(Float16, _tdistinvcdf(Float64(ν), Float64(p)))
+_tdistinvcdf(ν::Float32, p::Float32) = convert(Float32, _tdistinvcdf(Float64(ν), Float64(p)))
+tdistinvcdf(ν::Real, p::Real) = _tdistinvcdf(map(float, promote(ν, p))...)
 
 tdistinvccdf(ν::Real, p::Real) = -tdistinvcdf(ν, p)
 function tdistinvlogcdf(ν::T, logp::T) where {T <: Real}
```
devmotion, line 173: Should `tdistinvlogcdf` go through the new kernel as well? Right now `tdistinvcdf(0.5, 1e-8)` is fixed but `tdistinvlogcdf(0.5, log(1e-8))` still returns `-Inf`, and for `(2.5, log(1e-12))` the two disagree by 7e-9. The tail start is computed in log space anyway, so it looks like `log(pr)` could just become `logp` there. Happy for it to be a follow-up.

test/misc.jl
```diff
@@ -98,3 +98,78 @@
...
+                (5.0, 0.95, 2.0150483733330233),
+                (12.0, 0.3, -0.53861766820191637),
+                (50.0, 0.999, 3.261409055798318),
+                (300.0, 0.025, -1.9679030112610869),
+                (1000.0, 0.975, 1.9623390808264081),
+                (1.0e4, 1.0e-20, -9.282474153254304),
+                (1.0e6, 1 - 1.0e-12, 7.0345756932732169),
+                # the density underflows at these two points, exercising the
+                # early exit from the Newton polish
+                (0.9, 1.0e-150, -1.2783541486158767e166),
```
devmotion, line 125: At both of these `k2` underflows to exactly zero, so they exit through `f == 0`. The cases that go wrong are one decade earlier, where `k2` is subnormal but nonzero - eg `(0.9, 1e-144)`, `(0.35, 1e-56)` or `(0.7, 1e-110)`, see my comment in `tdist.jl`. Could we add one of those? Nothing bounds the iteration count either, so the points that need 100 iterations pass silently.

test/misc.jl
```diff
@@ -98,3 +98,78 @@
...
+                (1.5, 1.0e-200, -1.1245005997832135e133),
+            )
+            @test tdistinvcdf(ν, p) ≈ t rtol = 1.0e-13
+            @test tdistinvccdf(ν, p) == -tdistinvcdf(ν, p)
```
devmotion, line 129: `tdistinvccdf` is defined as `-tdistinvcdf`, so this holds by construction - is it testing anything?

test/misc.jl
```diff
@@ -98,3 +98,78 @@
...
+        end
+    end
+
+    @testset "round trips" begin
```
devmotion, line 133: The round trips and the Rmath tests both stay away from `p ≈ 1/2` - the Rmath grid only reaches `p = 0.5 ± 0.03`, and `rmathcomp` compares quantiles with `atol = 2e-14` - so nothing here notices the loss of relative accuracy near the median. Would a couple of reference values at `p = 0.5 - 1e-6` and similar be worth adding?

devmotion (commented):
I used Claude to check this against a 512-bit MPFR reference (own continued fraction for the incomplete beta, cross-checked against the closed forms for ν = 1 and ν = 2), and the numbers in the description reproduce for ν ≥ 1: the 15 reference values in the tests are all correct to ~1e-16, and over `p ∈ [1e-300, 1/2]` the maximum is 5.7e-14. Nice improvement.

Two things I ran into that I think are worth fixing, both commented inline. For `ν < 1` there is a window of `p` where `beta_inc` is called with a subnormal argument, and the polish then makes the result much worse than the starting approximation it was given - up to 34% relative error, and in some cases 100 iterations, ie ~11 µs. And relative accuracy near the median is worse than what we have today.

On the remaining gap to `qt` for ν = 5: as far as I can tell it is not the number of iterations. Both implementations evaluate the cdf/pdf pair twice, and ours is the cheaper of the two, 76 ns for `_tdistcdf_pdf` against 84 ns for `pt` + `dt` here. The difference is almost exactly `logbeta(ν / 2, 1/2)` + `exp`, which Hill's algorithm does not need at all: 334 ns for the kernel, 270 ns with `B` passed in, and `qt` is 270 ns. So there are two levers if we care. `B` only scales the Newton step, so a few digits would do there (full precision is only needed for the tail start, where it enters as `(ν B p)^(-1/ν)`), and the second `beta_inc` at ν = 5 is only there because of the 64x margin - one step from the CF start already gives 1.0e-14 and the quadratic estimate predicts that well, 6.1e-15. A Halley step does not help, it gives 2.0e-14 from the same start.

PR #231, opened 2026-09-21
Differentiate the gamma tail functions under ForwardDiff

### LINKED ISSUE #161: Relax _gammalogccdf argument types not to be restricted to Float64
I am getting the following error when I am trying to use ForwardDiff with Turing to sample Truncated Poisson distribution.

```julia
ERROR: MethodError: no method matching _gammalogccdf(::ForwardDiff.Dual{ForwardDiff.Tag{Turing.TuringTag, Float64}, Float64, 8}, ::ForwardDiff.Dual{ForwardDiff.Tag{Turing.TuringTag, Float64}, Float64, 8}, ::ForwardDiff.Dual{ForwardDiff.Tag{Turing.TuringTag, Float64}, Float64, 8})
Stacktrace:
  [1] gammalogccdf(k::Int64, θ::Int64, x::ForwardDiff.Dual{ForwardDiff.Tag{Turing.TuringTag, Float64}, Float64, 8})
  ```
  
 Which suggests that there is a problem with dispatch and perhaps types shouldn't be restricted to Float64.

### PR DESCRIPTION
**Pull request by ParadaCarleton**
Closes #161.

`SpecialFunctions.gamma_inc`, which every gamma tail bottoms out in, is defined for `Float16`, `Float32` and `Float64`, so `gammacdf`, `gammaccdf`, `_gammalogcdf` and `_gammalogccdf` throw on a `Dual`.

This adds a `ForwardDiff` extension that evaluates the tail on the primal and attaches the tangent analytically. Writing the tail as `P(k, x/θ)` gives `∂/∂θ = -(x/θ) ∂/∂x`, so one derivative carries both tangents, and the derivative itself is the gamma density already available as `gammalogpdf`.

A nonzero shape tangent throws rather than returning a wrong number: `∂/∂k` of the regularized incomplete gamma has no elementary closed form.

`test/forwarddiff.jl` checks the value and both tangents against finite differences for all four functions, including the complementary and log forms.

The two `test/qa.jl` ignores are needed because ForwardDiff declares no part of its dual-number API public, so ExplicitImports flags `Dual`, `value` and `partials`.

### CODE REVIEW

File: ext/StatsFunsForwardDiffExt.jl [hunk 1]
```diff
@@ -0,0 +1,66 @@
+module StatsFunsForwardDiffExt
```
devmotion: Convention nowadays is to not repeat the name of the main package in the extension name:
```suggestion
module ForwardDiffExt
```

ext/StatsFunsForwardDiffExt.jl
```diff
@@ -0,0 +1,66 @@
...
+    θ::Dual{T},
+    x::Dual{T},
+) where {T}
+    if !iszero(ForwardDiff.partials(k))
+        throw(ArgumentError(
+            "the gamma tail functions are not differentiable with respect to " *
+                "the shape parameter `k`: ∂/∂k of the regularized incomplete " *
+                "gamma has no elementary closed form",
+        ))
+    end
```
devmotion: I think this function should never receive a dual number for k but only operate on primal values for `k`.

ext/StatsFunsForwardDiffExt.jl
```diff
@@ -0,0 +1,66 @@
...
+    value_θ, value_x = ForwardDiff.value(θ), ForwardDiff.value(x)
```
devmotion: `ForwardDiff.value` (and `ForwardDiff.partials`) should never be called without tag (they are also internal but probably can't be avoided).

ext/StatsFunsForwardDiffExt.jl
```diff
@@ -0,0 +1,66 @@
...
+    tangent = derivative * ForwardDiff.partials(x) -
+        (derivative * value_x / value_θ) * ForwardDiff.partials(θ)
```
devmotion: This is not NaN-safe.

ext/StatsFunsForwardDiffExt.jl
```diff
@@ -0,0 +1,66 @@
...
+        :(exp($logdensity - primal))
+    else
+        :(exp($logdensity))
+    end
+    derivative = if complementary
+        :(-$magnitude)
+    else
+        magnitude
+    end
+    @eval function StatsFuns.$f(k::Dual{T}, θ::Dual{T}, x::Dual{T}) where {T}
```
devmotion: This is generally not general enough - one might not want to differentiate wrt to all parameters, and in fact one must not differentiate wrt to `k`.

One would generally want to - as for the DiffRules-based definitions in ForwardDiff - optimize this for different argument (also non-dual and differently tagged Dual) argument types.

test/qa.jl
devmotion, line 1: These changes must be reverted

ParadaCarleton:
Thanks. All six addressed in a42e542.

- **Module/file naming.** Fixed.
- **Generality / tags / NaN-safety.** Rewritten on `ForwardDiff.@define_ternary_dual_op`, so every argument/tag combination is generated, `value`/`partials` are called tag-qualified, and the tangents go through `dual_definition_retval`, i.e. `_mul_partials` and the NaN-safe path. Each branch only forms the partials whose argument is actually dual.
- **No differentiation wrt `k`.** The methods now attach to the *public* `gammalogcdf`/`gammalogccdf` rather than the `_`-prefixed internals: `gammalogcdf(k, θ, x) = _gammalogcdf(map(float, promote(k, θ, x))...)` promotes `k` to a `Dual` before the internal is reached, so a rule on the internal can never see a non-dual `k`. Pre-promotion, a dual `k` means the caller really is differentiating wrt the shape, and that throws.
- **test/qa.jl.** Reverted, except for four names on `check_all_qualified_accesses_are_public`: `value`, `partials`, `dual_definition_retval`, `@define_ternary_dual_op`, since ForwardDiff declares nothing public.

